PR #2, opened 2015-11-27
WIP added more quaternion methods

tumbili:
@yyassa This is the PR for the required quaternion methods. We need to provide 100 % test coverage to bring this in!
In the diff you can see that I did one example for a test.

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "math.hpp"
+#include "helper_functions.hpp"
 
 namespace matrix
 {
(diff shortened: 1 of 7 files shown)


tumbili:
@yyassa I implemented almost all tests, just one function missing.

jgoppert:
Great work! One consideration is that axis angle could derive from vector 3. This would be more aligned with the conversion in the rest of the library:

For instance, this is what I would expect the conversion from euler to axis-angle would look like to the end user.
```c++
Euler euler(1,2,3);
AxisAngle axisAngle(euler); 
Vector3 axis = axisAngle.axis();
float angle = axisAngle.angle();

AxisAngle axisAngle2(10, 0, 0); // angle 10, axis (1,0,0);
```

tumbili:
@jgoppert That's a good idea. Let me finish this up and after that I can add the axis angle class,

matrix/Quaternion.hpp
```diff
@@ -120,6 +121,69 @@ class Quaternion : public Vector<Type, 4>
         v(3) = w(2,0);
         return Q * v * Type(0.5);
     }
+
+    void invert() {
```
thomasgubler: maybe we should clearly separate the conjugate and the inverse
http://www.3dgep.com/understanding-quaternions/#Quaternion_Conjugate and http://www.3dgep.com/understanding-quaternions/#Quaternion_Inverse

jgoppert:
Manually merged.

PR #3, opened 2015-11-28
Int to size_t

yyassa:
More generalizable, fixes the sign comparison bug


yyassa:
Wrong base for PR, will re-open

PR #10, opened 2016-01-14
fix euler calculation

tumbili:
I didn't really look much into it but using the old calculation QGC only displays yaw from west over north to east. So it seems like yaw is in the range [-pi/2, pi/2].

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,10 @@ sudo: false
 install:
   - pip install --user cpp-coveralls
 script:
-  - cmake -DCMAKE_BUILD_TYPE=Profile .
+  - cmake -DCMAKE_BUILD_TYPE=Profile -DTEST=ON -DFORMAT=ON .
   - make
   - make check_format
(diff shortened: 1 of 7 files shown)


tumbili:
@jgoppert Could you have a look at this?

jgoppert:
@tumbili Your calculations were failing the unit test that converted from euler to dcm and back at gimbal lock. I fixed it by deriving the relationship with the dcm at gimbal lock assuming roll is zero. See the attached ipython notebook (https://github.com/PX4/Matrix/blob/euler_fix/doc/euler_gimbal_lock.ipynb) , if you want to check my math. I'm not sure why atan2 was giving you quadrant issues, I don't see any difference between this code and the previous method as far as that, but let me know if you still see the bug. Maybe we need another unit test for quadrant checking if there is a bug.

jgoppert:
I now have a complete round trip unit check working that tests all quadrants, so I'm confident to merge this.

PR #12, opened 2016-01-24
Fix error in conversion from rotation matrix to Euler angles

priseborough:
This patch fixes errors in the calculation of both roll and yaw angles and uses the faster single precision trig functions

### DIFF
--- a/matrix/Euler.hpp
+++ b/matrix/Euler.hpp
@@ -48,23 +48,13 @@ public:
 
     Euler(const Dcm<Type> & dcm) : Vector<Type, 3>()
     {
-        Type psi_val = Type(atan(dcm(1, 0)/ dcm(0, 0)));
-        Type theta_val = Type(asin(-dcm(2,0)));
-        Type phi_val = Type(atan(dcm(2, 1)/ dcm(2, 2)));
+        Type psi_val = Type(atan2f(dcm(1, 0),dcm(0, 0)));
+        Type theta_val = Type(asinf(fminf(fmaxf(-dcm(2,0),-1.0f),1.0f)));
+        Type phi_val = Type(atan2f(dcm(2, 1),dcm(2, 2)));
 
-        // protection against NaN if dcm(0,0) or dcm(2,2) == 0
-        psi() = 0;
-        theta() = 0;
-        phi() = 0;
-        if (psi() >= -(Type)M_PI_2 && psi() <= (Type)M_PI_2) {
-            psi() = psi_val;
-        }
-        if (theta() >= -(Type)M_PI_2 && theta() <= (Type)M_PI_2) {
-            theta() = theta_val;
-        }
-        if (phi() >= -(Type)M_PI_2 && phi() <= (Type)M_PI_2) {
-            phi() = phi_val;
-        }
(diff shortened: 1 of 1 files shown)


jgoppert:
@priseborough  This fails the unit test, people seem to be having trouble using the unit tests so I just spent some time to improve the output. Before committing, please run the unit tests locally:

```bash
mkdir build
cd build
cmake ..
make
ctest -V
```

```c++
11: test failed on /home/jgoppert/git/Matrix/test/attitude.cpp:96
11: not equal
11: x:
11: [[	         0],
11: [	    1.5708],
11: [	       0.2]]
11: y:
11: [[	  -3.14159],
11: [	    1.5708],
11: [	  -2.94159]]
```
This test is basically checking conversion from euler to dcm and back, so that equation doesn't work as is.

If you pull master and apply your patch, you should get the same results. There is already a patch waiting for review for this here, that passes the above test, https://github.com/PX4/Matrix/pull/10. Do you disagree with the math there?

matrix/Euler.hpp
```diff
@@ -48,23 +48,13 @@ class Euler : public Vector<Type, 3>
...
-            psi() = psi_val;
-        }
-        if (theta() >= -(Type)M_PI_2 && theta() <= (Type)M_PI_2) {
-            theta() = theta_val;
-        }
-        if (phi() >= -(Type)M_PI_2 && phi() <= (Type)M_PI_2) {
-            phi() = phi_val;
-        }
+        Type psi_val = Type(atan2f(dcm(1, 0),dcm(0, 0)));
+        Type theta_val = Type(asinf(fminf(fmaxf(-dcm(2,0),-1.0f),1.0f)));
```
jgoppert: so this looks like you are bounding the dcm value between -1 and 1, is this typically a problem?

matrix/Euler.hpp
```diff
@@ -48,23 +48,13 @@ class Euler : public Vector<Type, 3>
 
     Euler(const Dcm<Type> & dcm) : Vector<Type, 3>()
```
jgoppert: Maybe we need to have a single a double precision implementation.
Since the type for the dcm is either going to be float or double, this should be pretty easy to do with the templates.

jgoppert:
Closing this as I already addressed it in #10.

priseborough:
The maths for phi is still wrong and yes you should be bounding the input to asin rather than handling the exception after the NaN has been created.

Phi should be in the range from -pi to +pi

priseborough:
A single precision method is important - these trig functions are very expensive on the pixhawk in double precision.

priseborough:
Ignore the last comment about the maths, my git submodule had not updated properly.

jgoppert:
Ok, so the 2 things left to add are:

1. bounding asin input
2. creating single and double precision version of functions

If you want to create a short pull request I'll get that in.

jgoppert:
To make sure the euler angles were working for all quadrants I added a loop through all roll pitch yaw combos with a resolution of 90 degrees in the unit test and made sure I could recover the original euler angles from the dcm matrix. So I think that should help keep us from breaking anything going forward.

PR #14, opened 2016-02-25
Matrix index correction to allow non-square matrices to evaluate properly.

mcsauder:
This PR is to correct what appears to be a few index copy/paste errors that would function properly on square matrices but likely not in other cases.


jgoppert:
Thanks for catching that.

PR #15, opened 2016-04-21
Features and fixes

MaEtUgR:
all nice separated by descriptive commits
thanks in advance

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -497,8 +497,6 @@ std::ostream& operator<<(std::ostream& os,
 }
 #endif // defined(SUPPORT_STDIOSTREAM)
 
-typedef Matrix<float, 3, 3> Matrix3f;
-
 } // namespace matrix
 
 /* vim: set et fenc=utf-8 ff=unix sts=0 sw=4 ts=4 : */
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -274,28 +274,14 @@ public:
 
     void write_string(char * buf, size_t n) const
     {
+        buf[0] = '\0'; // make an empty string to begin with (we need the '\0' for strlen to work)
         const Matrix<Type, M, N> &self = *this;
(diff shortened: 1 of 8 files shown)


LorenzMeier:
Could you add the required unit tests so we don't loose test coverage? Thanks!

MaEtUgR:
on my way, hope i'll figure it out

MaEtUgR:
I figured it out so I can also write a test for the missing quaternion constructor... just a moment

MaEtUgR:
done

LorenzMeier:
Awesome, thanks!

LorenzMeier:
@jgoppert Can you review?

matrix/Matrix.hpp
```diff
@@ -274,28 +274,14 @@ class Matrix
 
     void write_string(char * buf, size_t n) const
     {
+        buf[0] = '\0'; // make an empty string to begin with (we need the '\0' for strlen to work)
```
jgoppert: Has this function been tested on nuttx?

Also, what about buffer overflow concerns with this approach? I think a fixed buffer and snprintf are better, but you can still use your buffer appending to prevent all of the calls to snprintf.

jgoppert:
If you want to put print changes in a separate pr we can get the rest merged now.

matrix/Matrix.hpp
MaEtUgR: i tested the print on nuttx but buffer overflow is a good point and got lost on my todo list, I'm solving it now and will report

MaEtUgR:
sorry for not being available friday and weekend
i prvented possible buffer overflows now with snprintf, thanks for reminding me
i tested it in the ubuntu sitl environment and a 3DR PX4 Board on nuttx

MaEtUgR:
@jgoppert is it ok now? or do I still miss something?

jgoppert:
Thanks! Looks good.

PR #16, opened 2016-05-03
WIP: refactoring rotation classes

birchera:
* Refactoring dcm, euler and quaternion classes. Code to initialize instances has been factored out in separate functions to allow assignment to already initialized instances.
* Adding initial description, will elaborate on that.

Compiles and flies in SITL. Further testing will be necessary.

### DIFF
--- a/matrix/Dcm.hpp
+++ b/matrix/Dcm.hpp
@@ -46,7 +46,7 @@ public:
     /**
      * Constructor from array
      *
-     * Copies the given scalar value to all matrix fields.
+     * More elaborate function description
      *
      * @param other array
      */
--- a/matrix/Dcm.hpp
+++ b/matrix/Dcm.hpp
@@ -23,7 +23,8 @@ class Euler;
 /**
  * Direction cosine matrix class
  *
- * More elaborate class description
+ * The rotation between two coordinate frames is
+ * described by this class.
  */
 template<typename Type>
 class Dcm : public Matrix<Type, 3, 3>
@@ -46,8 +47,6 @@ public:
     /**
      * Constructor from array
      *
-     * More elaborate function description
-     *
      * @param other array
      */
     Dcm(const Type *data_) : Matrix<Type, 3, 3>(data_)
@@ -56,8 +55,6 @@ public:
 
     /**
      * Copy constructor
-     *
-     * More elaborate function description
      *
      * @param other Matrix33 to set dcm to
      */
@@ -68,7 +65,8 @@ public:
     /**
      * Constructor from quaternion
      *
-     * More elaborate function description
+     * Instance is initialized from quaternion representing
+     * transformation from inertial frame to body frame.
      *
      * @param q quaternion to set dcm to
      */
@@ -79,7 +77,8 @@ public:
     /**
      * Constructor from euler angles
      *
-     * More elaborate function description
+     * Instance is initialized from angle tripplet (3,2,1) representing
+     * transformation from body frame to inertial frame. 
      *
      * @param euler euler angles to set dcm to
      */
@@ -91,7 +90,8 @@ public:
     /**
      * Set from quaternion
      *
-     * More elaborate function description
+     * Instance is set from quaternion representing
+     * transformation from inertial frame to body frame.
      *
      * @param q quaternion to set dcm to
      */
@@ -119,7 +119,8 @@ public:
     /**
      * Set from euler angles
      *
-     * More elaborate function description
+     * Instance is set from angle tripplet (3,2,1) representing
+     * transformation from body frame to inertial frame. 
      *
      * @param euler euler angles to set dcm to
      */
--- a/matrix/Dcm.hpp
+++ b/matrix/Dcm.hpp
@@ -20,6 +20,11 @@ class Quaternion;
 template<typename Type>
 class Euler;
 
+/**
+ * Direction cosine matrix class
+ *
+ * More elaborate class description
+ */
 template<typename Type>
 class Dcm : public Matrix<Type, 3, 3>
 {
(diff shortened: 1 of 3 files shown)


matrix/Dcm.hpp
```diff
@@ -28,20 +34,68 @@ class Dcm : public Matrix<Type, 3, 3>
...
+    Dcm(const Euler<Type> & euler) {
+        set_from_euler(euler);
+    }
+
+
+    /**
+     * Set from quaternion
+     *
+     * Instance is set from quaternion representing
+     * transformation from inertial frame to body frame.
```
priseborough: The terms earth and body are very application specific. I recommend referring the starting frame as 'frame 1' and the finishing frame as 'frame 2'

matrix/Dcm.hpp
```diff
@@ -20,6 +20,12 @@ class Quaternion;
 template<typename Type>
 class Euler;
 
+/**
+ * Direction cosine matrix class
+ *
+ * The rotation between two coordinate frames is
```
priseborough: Does the DCm matrix rotate from frame 2 to frame 1 or vice versa?

matrix/Dcm.hpp
```diff
@@ -28,20 +34,68 @@ class Dcm : public Matrix<Type, 3, 3>
...
+     * @param q quaternion to set dcm to
+     */
     Dcm(const Quaternion<Type> & q) {
+        set_from_quaternion(q);
+    }
+
+    /**
+     * Constructor from euler angles
+     *
+     * Instance is initialized from angle tripplet (3,2,1) representing
```
priseborough: you mean a rotation sequence about the 3rd, 2nd and 1st axis in that order?

matrix/Dcm.hpp
```diff
@@ -28,20 +34,68 @@ class Dcm : public Matrix<Type, 3, 3>
...
+     * transformation from body frame to inertial frame.
```
priseborough: Avoid using application specific terms like 'earth' and 'body'

matrix/Euler.hpp
```diff
@@ -19,34 +19,113 @@ class Dcm;
 template <typename Type>
 class Quaternion;
 
+/**
+ * Euler angles class
+ *
+ * This class describes the transformation from the body fixed frame
+ * to the inertial frame via 3-2-1 tait brian euler angles.
```
priseborough: Tait-Bran sequences can be intrinsic or extrinsic. An unambiguous description would be:

This class describes the rotation from the frame 1 to frame 2 via a 3-2-1 intrinsic Tait-Bryan rotation sequence

matrix/Dcm.hpp
```diff
@@ -62,7 +116,15 @@ class Dcm : public Matrix<Type, 3, 3>
         dcm(2, 2) = aSq - bSq - cSq + dSq;
     }
 
-    Dcm(const Euler<Type> & euler) {
+    /**
+     * Set from euler angles
+     *
+     * Instance is set from angle tripplet (3,2,1) representing
```
priseborough: This provides the transformation matrix from frame 2 to frame 1 where the rotation from frame 1 to frame 2 is described by a 3-2-1 intrinsic Tait-Bryan rotation sequence

priseborough:
Some general comments:

A good interface definition will answer the following questions: 

1) How are the start and end frames in the rotation sequence defined?, (a consistent lexicon should be used, eg frame 1 and frame 2)

2) Where rotation sequences are used, what is the start and finish frame, what is the axis rotation order, are intrinsic or extrinsic rotations used?. 

3) Where quaternions or transformation matrices are used, what is the starting and finishing frame?

4) Do all rotations follow the RH rule?

5) What angular units are used to describe rotations (rad, deg?)

6) Is the scalar part of the quaternion contained in the first or last element in the vector?

matrix/Dcm.hpp
tumbili: @priseborough It rotates from frame 1 into frame 2 not like I said before.

matrix/Dcm.hpp
tumbili: @priseborough yes, intrinsic rotation about z, y and then x. Will document this better.

matrix/Euler.hpp
tumbili: @priseborough agreed on intrinsic rotations since this is most convenient for UAVs

tumbili:
@birchera I will wrap this up considering the comments from Paul

jgoppert:
@birchera Can you explain why you added the set_from_euler functions etc? This is already handled by the constructors more cleanly.

```c++
Quaternionf q(1,0,0,0);
Quaternionf q2(1,0,0,0);
Eulerf e(q);
e = Eulerf(q2);
```

vs.

```c++
Quaternionf q(1,0,0,0);
Quaternionf q2(1,0,0,0);
Eulerf e(q);
e = q2.from_euler(e);
```

I was trying to keep to just constructors to make the conversions implicit as much as possible, but I'd like to hear what everyone prefers, and if there are arguments for providing these other functions. I think one drawback of having multiple ways of coding the same thing is that we will have people writing the same code in many different styles.

matrix/Dcm.hpp
jgoppert: Maybe we should say that for uav applications frame 1 is earth and frame 2 is body. The Matrix library only supports one axis angle sequence, and the point is that this makes it clean and you can easily tranform between dcm, euler, and quaternions without worrying about it. I don't think we should ever support more than one euler sequence.

priseborough: Something to consider is that most 3-axis gimbal use a 312 intrinsic rotation from body to sensor frame and we had to implement a 312 method  in APM when we added support for gimbal stabilisation.

jgoppert:
@priseborough If there is enough demand for other euler sequences I think we could make the sequence a template parameter and then implement the conversions for specific sequences.

tumbili:
@jgoppert The goal was to have one function which does the conversion and the also use it in the constructor. Like this there is no code duplication. At the moment it's a pain in the ass if you need to convert e.g. a quaternion from an euler instance if the quaternion instance itself already exists.
That's why we want to split the functionality.

birchera:
Moved to https://github.com/PX4/Matrix/pull/19

PR #17, opened 2016-05-05
Matrix.setCol off by one bug fixed

MaEtUgR:
The Column Vector you are copying from, only has one Column and it is indexed by 0 not 1.
I also completed the unit test stub that would have found this bug.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -341,7 +341,7 @@ public:
     {
         Matrix<Type, M, N> &self = *this;
         for (size_t i = 0; i < M; i++) {
-            self(i, j) = col(i, 1);
+            self(i, j) = col(i, 0);
         }
     }
 
--- a/test/matrixAssignment.cpp
+++ b/test/matrixAssignment.cpp
@@ -78,8 +78,14 @@ int main()
     TEST(fabs(m5(0,0) - s) < 1e-5);
 
     Matrix<float, 2, 2> m6;
-    m6.setRow(0, Vector2f(1, 1));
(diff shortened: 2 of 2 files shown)


LorenzMeier:
Thanks! @jgoppert Are there more unit test stubs? I was surprised to find this one.

jgoppert:
@LorenzMeier  This wasn't exactly a test stub, just a test that I never thought of and a bug I never had run into. @MaEtUgR thanks for finding this.

I'm sure that there are a few more goofs, but I think with more users and more tests we will eventually find them.

jgoppert:
@LorenzMeier It is odd that the coverage testing didn't find this for us. We should look into that more.

MaEtUgR:
sry I just named it stub because there was already a call to this method in the test `m6.setCol(0, Vector2f(1, 1));` but the result was not checked afterwards with a TEST(...). I would say coverage just checks if it's called by a test, not if it's actually compared to a ground truth...

MaEtUgR:
to be honest I found it pretty fast because the gcc compiler (at least with the settings from pixhawk cmake) made a halt on warning for statically checked array index exceedance.

I guess the test just copiled with a warning and then wrote on some wrong stack memory, nothing was checked after that so it went just fine.

jgoppert:
@MaEtUgR didn't realize I had done that, yeah I would call that a stub too. That was likely my fault and I forgot to add the check.

MaEtUgR:
@jgoppert yea i thought it was something like that, nevermind, let's fix everything we find :)

PR #19, opened 2016-05-10
Pr matrix documentation

tumbili:


matrix/Dcm.hpp
```diff
@@ -28,20 +40,69 @@ class Dcm : public Matrix<Type, 3, 3>
...
     {
         (*this) = eye<Type, 3>();
     }
 
+    /**
+     * Constructor from array
+     *
+     * @param other array
+     */
     Dcm(const Type *data_) : Matrix<Type, 3, 3>(data_)
```
LorenzMeier: param here is data_

matrix/Euler.hpp
```diff
@@ -19,34 +24,117 @@ class Dcm;
...
     }
 
-    Euler(const Dcm<Type> & dcm) : Vector<Type, 3>()
+    /**
+     * Set from dcm
+     *
+     * Instance is set from dcm representing transformation
+     * from frame 2 to frame 1.
+     *
+     * @param dcm_ dcm to set angles to
```
LorenzMeier: parameter is dcm, not dcm_.

matrix/Quaternion.hpp
```diff
@@ -44,14 +58,72 @@ class Quaternion : public Vector<Type, 4>
...
+     * @param euler euler angle instance
+     */
+    Quaternion(const Euler<Type> & euler) :
+        Vector<Type, 4>()
+    {
+        set_from_euler(euler);
+    }
+
+    /**
+     * Constructor from quaternion values
```
LorenzMeier: The quaternion order is missing here - a good way to express this is to provide the zero-rotation / identify example, which is (1, 0, 0, 0).

matrix/Quaternion.hpp
```diff
@@ -83,8 +163,18 @@ class Quaternion : public Vector<Type, 4>
                sinPhi_2 * sinTheta_2 * cosPsi_2;
     }
 
-    Quaternion(Type a, Type b, Type c, Type d) :
-        Vector<Type, 4>()
+    /**
+     * Set from quaternion values
+     *
+     * Instance is set from quaternion values representing coordinate
```
LorenzMeier: Quaternion order missing here.

tumbili: @LorenzMeier I've added it here. Additionally, the quaternion order is specified in the description at the top.

tumbili:
@jgoppert In the old implementation how was it possible to set an existing quaternion instance to an existing euler angle instance? Since the conversions were only done in the constructor I've found no way to this without creating another quaternion instance.

tumbili:
@jgoppert Actually I wasn't aware of these implicit conversions. You are right, it's actually much cleaner. I will revert this back.

matrix/Dcm.hpp
```diff
@@ -3,6 +3,12 @@
  *
  * A direction cosine matrix class.
  *
+ * This library uses the convention that premultiplying a three dimensional
+ * vector represented in coordinate system 1 will apply a rotation from coordinate system
+ * 1 to coordinate system 2 to the vector. The rotation is performed following the right-hand rule.
```
priseborough: "All rotations and axis systems follow the right-hand rule"

matrix/Dcm.hpp
```diff
@@ -28,20 +40,69 @@ class Dcm : public Matrix<Type, 3, 3>
...
+    Dcm(const Euler<Type> & euler) {
+        set_from_euler(euler);
+    }
+
+
+    /**
+     * Set from quaternion
+     *
+     * Instance is set from quaternion representing
+     * transformation from frame 2 to frame 1.
```
priseborough: Ae you sure about this? I thought the quaternion was from 1 to 2

matrix/Euler.hpp
```diff
@@ -19,34 +24,117 @@ class Dcm;
 template <typename Type>
 class Quaternion;
 
+/**
+ * Euler angles class
+ *
+ * This class describes the transformation from the body fixed frame
```
priseborough: replace inertial and body with frame 1 and frame 2

matrix/Euler.hpp
```diff
@@ -19,34 +24,117 @@ class Dcm;
...
+     */
     Euler(const Matrix<Type, 3, 1> & other) :
         Vector<Type, 3>(other)
     {
     }
 
+    /**
+     * Constructor from euler angles
+     *
+     * Instance is initialized from angle tripplet (3,2,1)
```
priseborough: Lexicon is not consistent with previous usage

matrix/Dcm.hpp
tumbili: @priseborough I will add a conjugate method and try it out on a vector. How would you find out else?

birchera: I agree that it should be from 1 to 2. Thanks for pointing out the mistake I did in the initial description.

matrix/Euler.hpp
```diff
@@ -66,8 +156,15 @@ class Euler : public Vector<Type, 3>
         psi() = psi_val;
     }
 
-    Euler(const Quaternion<Type> & q) :
-        Vector<Type, 3>()
+    /**
+     * Set from dcm
+     *
+     * Instance is set from quaternion representing
+     * transformation from body frame to inertial frame.
```
birchera: from frame 2 to frame 1

matrix/Quaternion.hpp
```diff
@@ -44,14 +63,73 @@ class Quaternion : public Vector<Type, 4>
...
+        Vector<Type, 4>()
+    {
+        set_from_euler(euler);
+    }
+
+    /**
+     * Constructor from quaternion values
+     *
+     * Instance is initialized from quaternion values representing coordinate
+     * transformation from frame 2 to frame 1.
```
birchera: why from frame 2 to frame 1? To be consistent it should be 1 to 2.

I see the problem that we still try to explain how the frames are used within the px4 code without calling them by the real names (inertial/body). This only leads to confusion.

The reason this seems necessary is that euler angles describe the reverse transformation from the other representations and frames are switched in the conversion functions.

I wonder if it would make sense to redefine the transformation of dcm and quaternions to have consistent rotation representations?

tumbili:
@birchera I reverted the conversion methods back to go into the constructor.
Also I provided some more description for the quaternion class, e.g which formula to use to rotate a vector and how to concatenate rotations.

tumbili:
@birchera @priseborough Is there something missing here? I added a small description at the top of the quaternion class which states the quaternion rotation formula and the order of how to concatenate quaternions.

tumbili:
@LorenzMeier @birchera I'm going to merge this since I consider it to be complete and it's just documentation anyway.

birchera:
I am fine with merging, looks good to me.

Issue #18, opened 2016-05-09
Add class for axis angle representation


jgoppert:
This code breaks the style of the rest of the library:

```c++
    // get rotation axis from quaternion (nonzero rotation)
    q = Quatf(cosf(1.0f / 2), 0.0f, sinf(1.0f / 2), 0.0f);
    rot = q.to_axis_angle();
    TEST(fabsf(rot(0)) < eps);
    TEST(fabsf(rot(1) -1.0f) < eps);
    TEST(fabsf(rot(2)) < eps);

    // get rotation axis from quaternion (zero rotation)
    q = Quatf(1.0f, 0.0f, 0.0f, 0.0f);
    rot = q.to_axis_angle();
    TEST(fabsf(rot(0)) < eps);
    TEST(fabsf(rot(1)) < eps);
    TEST(fabsf(rot(2)) < eps);

    // from axis angle (zero rotation)
    rot(0) = rot(1) = rot(2) = 0.0f;
    q.from_axis_angle(rot, 0.0f);
    q_true = Quatf(1.0f, 0.0f, 0.0f, 0.0f);
    TEST(fabsf(q(0) - q_true(0)) < eps);
    TEST(fabsf(q(1) - q_true(1)) < eps);
    TEST(fabsf(q(2) - q_true(2)) < eps);
    TEST(fabsf(q(3) - q_true(3)) < eps);


```

PR #25, opened 2016-07-06
Axis angle

jgoppert:
This closes https://github.com/PX4/Matrix/issues/18

@tumbili @bkueng @LorenzMeier 

to_axis_angle, from_axis_angle should now be deprecated.

### DIFF
--- a/matrix/AxisAngle.hpp
+++ b/matrix/AxisAngle.hpp
@@ -1,0 +1,158 @@
+/**
+ * @file AxisAngle.hpp
+ *
+ * @author James Goppert <james.goppert@gmail.com>
+ */
+
+#pragma once
+
+#include "math.hpp"
+#include "helper_functions.hpp"
+
+namespace matrix
+{
+
+template <typename Type>
+class Dcm;
+
+template <typename Type>
+class Euler;
+
+template <typename Type>
+class AxisAngle;
+
+/**
+ * AxisAngle class
+ *
+ * The rotation between two coordinate frames is
+ * described by this class.
+ */
+template<typename Type>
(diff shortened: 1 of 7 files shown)


matrix/AxisAngle.hpp
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @param q quaternion
+     */
+    AxisAngle(const Quaternion<Type> &q) :
+        Vector<Type, 3>()
+    {
+        AxisAngle &v = *this;
+        Type angle = (Type)2.0f*acosf(q(0));
+        Type mag = sinf(angle/2.0f);
+        v(0) = angle*q(1)/mag;
```
bkueng: If `angle` = 0, you have a division by 0

matrix/AxisAngle.hpp
```diff
@@ -0,0 +1,158 @@
...
+     * Instance is initialized from a dcm representing coordinate transformation
+     * from frame 2 to frame 1.
+     *
+     * @param dcm dcm to set quaternion to
+     */
+    AxisAngle(const Dcm<Type> &dcm) :
+        Vector<Type, 3>()
+    {
+        AxisAngle &v = *this;
+        v = Quaternion<float>(dcm);
```
bkueng: I think this should be `Type` instead of float

matrix/AxisAngle.hpp
jgoppert: Thanks for the catch.

matrix/AxisAngle.hpp
jgoppert: Right, will fix this as well.

PR #26, opened 2016-08-01
add support for inversed() on nonsquare matrices as convenience. 

mkschreder:
allows using inversed() directly on an instance of Matrix class, provided that it is a square matrix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ set(CMAKE_CXX_FLAGS
 	-Werror=uninitialized
 	-Werror=init-self
 	-Wcast-qual
-	-Wconversion
+	-Wno-conversion
 	-Wcast-align
 	-Werror
 	)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,7 @@ set(CMAKE_CXX_FLAGS
 	-Werror=uninitialized
 	-Werror=init-self
 	-Wcast-qual
-	-Wno-conversion
+	-Wconversion
 	-Wcast-align
 	-Werror
 	)
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -13,6 +13,16 @@
  * represents the real part, thus, a quaternion representing a zero-rotation
  * is defined as (1,0,0,0).
  *
+ * Using Hamilton Quaternion
+ *  - https://en.wikipedia.org/wiki/Quaternion
+ *  - http://www.iri.upc.edu/people/jsola/JoanSola/objectes/notes/kinematics.pdf
+ *  - Components (q_w, q_v) where q_w is the scalar and q_v is a vector
+ *  - algebrea: i*j = k
+ *  - handedness: right
+ *  - function: passive (rotation operator rotates frames)
+ *  - right-to-left product means local-to-global, q_GL
+ *    default operator v_G = q x v_L x q*
+ *
  * @author James Goppert <james.goppert@gmail.com>
  */
 
@@ -190,9 +200,9 @@ public:
         const Quaternion &p = *this;
         Quaternion r;
         r(0) = p(0) * q(0) - p(1) * q(1) - p(2) * q(2) - p(3) * q(3);
-        r(1) = p(0) * q(1) + p(1) * q(0) - p(2) * q(3) + p(3) * q(2);
-        r(2) = p(0) * q(2) + p(1) * q(3) + p(2) * q(0) - p(3) * q(1);
-        r(3) = p(0) * q(3) - p(1) * q(2) + p(2) * q(1) + p(3) * q(0);
+        r(1) = p(0) * q(1) + p(1) * q(0) + p(2) * q(3) - p(3) * q(2);
+        r(2) = p(0) * q(2) - p(1) * q(3) + p(2) * q(0) + p(3) * q(1);
+        r(3) = p(0) * q(3) + p(1) * q(2) - p(2) * q(1) + p(3) * q(0);
         return r;
     }
 
@@ -239,7 +249,7 @@ public:
     {
         const Quaternion &q = *this;
         Quaternion<Type> v(0, w(0, 0), w(1, 0), w(2, 0));
-        return v * q  * Type(0.5);
+        return q * v  * Type(0.5);
     }
 
     /**
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -238,19 +238,8 @@ public:
     Matrix41 derivative(const Matrix31 &w) const
     {
         const Quaternion &q = *this;
-        Type dataQ[] = {
-            q(0), -q(1), -q(2), -q(3),
-            q(1),  q(0), -q(3),  q(2),
-            q(2),  q(3),  q(0), -q(1),
-            q(3), -q(2),  q(1),  q(0)
-        };
-        Matrix<Type, 4, 4> Q(dataQ);
-        Vector<Type, 4> v;
-        v(0) = 0;
-        v(1) = w(0, 0);
-        v(2) = w(1, 0);
-        v(3) = w(2, 0);
-        return Q * v * Type(0.5);
+        Quaternion<Type> v(0, w(0, 0), w(1, 0), w(2, 0));
+        return v * q  * Type(0.5);
     }
 
     /**
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -44,6 +44,9 @@ template <typename Type>
 class AxisAngle;
 
 
+template<typename Type>
+Quaternion<Type> normalize(const Quaternion<Type> &o);
+
 /**
  * Quaternion class
  *
@@ -189,6 +192,10 @@ public:
         q(3) = d;
     }
 
+    Quaternion(const Vector<Type, 3> &axis, Type angle) {
+        from_axis_angle(axis, angle);
+    }
+
     /**
      * Quaternion multiplication operator
      *
@@ -206,12 +213,17 @@ public:
         return r;
     }
 
+    Quaternion operator+(const Quaternion &o) const {
+        const Quaternion &p = *this;
+        return Quaternion(p(0) + o(0), p(1) + o(1), p(2) + o(2), p(3) + o(3));
+    }
+
     /**
      * Self-multiplication operator
      *
      * @param other quaternion to multiply with
      */
-    void operator*=(const Quaternion &other)
+    void operator*=(const Quaternion & other)
     {
         Quaternion &self = *this;
         self = self * other;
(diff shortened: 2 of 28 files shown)


julianoes:
The indenting seems wrong. Functionally I have no opinion.

mkschreder:
In the long run I think it is a good idea to make this library follow the
convention of matlab so that code can be easily moved back and forth
between the two. It is very close but about 1% further than the old mathlib
was.

jgoppert:
@mkschreder Making it straight forward to use and similar to matlab is a major goal. Can you fix the indenting and add a unit test to cover this? Also a check to make sure it doesn't try to invert non-square shape would be great. I think that should be caught in the square matrix ctor luckily.

mkschreder:
Yes it should not compile for non-square. I have added many more additions
in my own branch. Will submit when I have time.

jgoppert:
@mkschreder This is how to do the formatting:

```bash
mkdir build
cd build
cmake .. -DFORMAT=ON
 make format
```

Also make check_format is helpful.

mkschreder:
I have added more stuff (quaternion lerp, fixed matrix setRow etc)

matrix/Vector3.hpp
```diff
@@ -83,6 +83,24 @@ class Vector3 : public Vector<Type, 3>
 
 typedef Vector3<float> Vector3f;
 
+template <typename Type>
+class Quaternion;
+
+template<typename Type>
+Vector3<Type> operator* (const Quaternion<Type> &q, const Vector3<Type>& v) {
```
jgoppert: Some more description of what this function is doing would be useful.

matrix/Quaternion.hpp
```diff
@@ -213,10 +225,10 @@ class Quaternion : public Vector<Type, 4>
      * @param scalar scalar to multiply with
```
jgoppert: Need to switch scalar to s.

matrix/Quaternion.hpp
```diff
@@ -254,23 +266,40 @@ class Quaternion : public Vector<Type, 4>
...
      * Invert quaternion
      */
-    void invert()
-    {
+    void invert() {
         Quaternion &q = *this;
         q(1) *= -1;
         q(2) *= -1;
         q(3) *= -1;
     }
```
jgoppert: dot operator should be inherited from vector, no need to redefine in quaternion

test/attitude.cpp
```diff
@@ -49,8 +49,8 @@ int main()
     TEST(isEqual(euler_copy, euler_check));
 
     // quaternion ctor
```
jgoppert: What is your reasoning for not using Quatf alias?

matrix/Vector3.hpp
mkschreder: I have changed that to be quaternion multiplication formula. I don't know if I previously got "wrong" results because of non hamilton representation of quaternion. I did not have chance to test the nvidia version with new updates yet. It was not completely wrong - just wrong in the context of this library. But for the sake of being really sure that I always get correct result I have changed it to be q(0, vnew) = q * q(0, v) * q.inversed();.

matrix/Quaternion.hpp
mkschreder: There was a compiler error associated with some use case where I was not able to write code without using a temporary value to store intermediate result. Therefore I added dot to quaternion and it fixed the compiler error. I will let you know exactly what the problem is next time I encounter it with another class. So this is purely for being able to write code in an intuitive way without getting compiler errors.

mkschreder: I think I had to cast some intermediate value explicitly to Quaternion because compiler was unable to convert Vector4 to quaternion automatically in that context.

test/attitude.cpp
mkschreder: Purely a preference. Nothing that I would expect anyone else to do.

mkschreder:
Also, I make some operators overloaded global because it avoids compiler errors in some corner cases that I come across. I will document that more next time I stumble upon it.

CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ set(CMAKE_CXX_FLAGS
 	-Werror=uninitialized
 	-Werror=init-self
 	-Wcast-qual
-	-Wconversion
+	-Wno-conversion
```
jgoppert: Why did you turn of conversion warning, this is important for the embedded sign to force user to realize if he is wasting casts.

test/test_macros.hpp
```diff
@@ -8,5 +8,22 @@
...
+inline float degrees(float rad) {
+    return rad / (M_PI / 180.0f);
+}
+
+inline bool is_equal(float a, float b, float e) {
+    return fabsf(a - b) < e;
+}
+inline bool is_equal(float a, float b) {
+    return is_equal(a, b, FLT_EPSILON);
+}
```
jgoppert: I just add a isEqual function to master, it should cover this use case.

mkschreder: Don't use float & in it though. Use just float. I have not checked if you
updated that yet.

CMakeLists.txt
mkschreder: I did not want to cast every single constant in my test code from double
(since my system uses doubles by default) to float. My reasoning is that
most embedded systems use float so unexpected automatic casts from double
to float are rarely going to be a problem in practice. I don't think extra
casts are ever a problem either because compiler should be smart enough to
optimize that.

test/test_macros.hpp
jgoppert: Yes, just passing by value.

CMakeLists.txt
jgoppert: We use the same compiler flags as px4 so not really appealing to change them. When you declare a constant, just put an f at the end and it will be a float constant instead of a double constant.

mkschreder: I thought that too but even 1.0f results in a complaint that I have an
implicit conversion from double to float and that I may be losing precision
from that (warning makes sense and is important but in that context it is
nonsense. Specially together with -Werror) When you build for px4 obviously
1.0f will be a float since that what is natively supported by hardware so
no problem there.

mkschreder: An alternative solution:
#pragma GCC diagnostic warning "-Wconversion"

I have reenabled Wconversion in cmake file.

mkschreder:
One good reason to move conversions from constructors into global functions (as I started doing in a few places) is that it forces one to explicitly state intent when coding. 

ie. if you have a matrix constructor that builds matrix from quaternion, say Dcm(quat), then it also implicitly allows one to write m = q where m can be called x or foo or whatever. This may be less typing but it also means code that is hard to read because you can not directly tell what the type of m variable is (since q is probably a local variable you probably know it is a quat but m is not a quat, yet you would assume it is until you look it up elsewhere in the code) If the conversion is called cosineMatrix(...) and has a bunch of overloads then you are always forced to type cosineMatrix when doing the conversions which makes it clear what the type of the result is. 

Another good reason for it is that it makes coding easier because you can make the code clear without having to know the template parameter of the target variable in the assignment. For instance if I have Euler<float> e and I want to convert a quat to it then instead of writing e = Euler<float>(q) (or e = q which is the confusing variant) I would write e = eulerAngles(q) where eulerAngles is a global template function that does the conversion. This both solves the ambiguity when trying to infer the type of e when reading code and also makes it clear that a conversion is taking place here. 

GLM library uses this paradigm a lot, producing highly readable code.

jgoppert:
@mkschreder You might want to shrink down and split up this PR, it is getting pretty large and hard to review.

I like the global template function approach, but I typically use the ctors to the same effect and I don't really find them less readable since I used the alias's.

```c++
    Dcmf dcm3(Eulerf(1, 2, 3));
    Dcmf dcm4(Eulerf(4, 5, 6));
    Dcmf dcm34 = dcm3*dcm4;
    TEST(isEqual(Eulerf(Quatf(dcm3)*Quatf(dcm4)), Eulerf(Quatf(dcm34))));
```

vs. 

```c++
    Dcmf dcm3(Eulerf(1, 2, 3));
    Dcmf dcm4(Eulerf(4, 5, 6));
    Dcmf dcm34 = dcm3*dcm4;
    TEST(isEqual(eulerAngles(quaternion(dcm3)*quaternion(dcm4)),
                eulerAngles(quaternion(dcm34))));
```

With regard to cross-type assignment, I think it can be a bit un-intuitive with poor variable names if used a lot, but I find it useful and clear with typical variable names.

```c++
Eulerf euler_nb(0, 0, 0);
Quatf quat_nb(1, 0, 0, 0);
euler_nb = quat_nb;
```

test/test_macros.hpp
```diff
@@ -11,19 +11,31 @@
...
+    return rad / (Type(M_PI) / Type(180.0));
 }
 
 inline bool is_equal(float a, float b, float e) {
     return fabsf(a - b) < e;
 }
+
 inline bool is_equal(float a, float b) {
     return is_equal(a, b, FLT_EPSILON);
 }
```
jgoppert: You should remove any non-template functions, when they get included multiple times in different object files it causes issues. Also, I think the function I added should work for this.

mkschreder:
The pull request is updated automaticalky. I just push updates to my repo.
It actually just started with a few commits in the begining..

mkschreder:
Ok. The only issue with that is that you have more places to change if you
one day decide to use doubles..

These last few things are just matter of taste so I don't expect anyone to
use one approach or the other. I think fixing functionality is most
important so we should primarily focus on that.

On 16 Aug 2016 21:43, "James Goppert" <notifications@github.com> wrote:

@mkschreder <https://github.com/mkschreder> You might want to shrink down
and split up this PR, it is getting pretty large and hard to review.

I like the global template function approach, but I typically use the ctors
to the same effect and I don't really find them less readable since I used
the alias's.

    Dcmf dcm3(Eulerf(1, 2, 3));
    Dcmf dcm4(Eulerf(4, 5, 6));
    Dcmf dcm34 = dcm3*dcm4;
    TEST(isEqual(Eulerf(Quatf(dcm3)*Quatf(dcm4)), Eulerf(Quatf(dcm34))));

vs.

    Dcmf dcm3(Eulerf(1, 2, 3));
    Dcmf dcm4(Eulerf(4, 5, 6));
    Dcmf dcm34 = dcm3*dcm4;
    TEST(isEqual(eulerAngles(quaternion(dcm3)*quaternion(dcm4)),
                eulerAngles(quaternion(dcm34))));

With regard to cross-type assignment, I think it can be a bit un-intuitive
with poor variable names if used a lot, but I find it useful and clear with
typical variable names.

Eulerf euler_nb(0, 0, 0);
Quatf quat_nb(1, 0, 0, 0);
euler_nb = quat_nb;

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/PX4/Matrix/pull/26#issuecomment-240215089>, or mute the
thread

test/test_macros.hpp
mkschreder: I changed them to be templated to get rid of some of the double to float
crap (Wconversion)..

jgoppert:
@mkschreder I realize the convenience of a devel branch, but probably best not to have it open on a PR. When you get another bug you want fixed please open a PR with a small fix I can review easily and I will get it in quickly.

PR #31, opened 2016-10-31
add DCM renormalize method

kd0aij:
@jgoppert It seemed like adding this method here made sense when I was rebasing https://github.com/PX4/Firmware/pull/5498 to deal with the elimination of Euler angles from the vehicle attitude setpoint topic. Let me know if you think it should go somewhere else.

### DIFF
--- a/matrix/Dcm.hpp
+++ b/matrix/Dcm.hpp
@@ -41,129 +41,134 @@ template<typename Type>
 class Dcm : public Matrix<Type, 3, 3>
 {
 public:
-    virtual ~Dcm() {};
+	virtual ~Dcm() {};
 
-    typedef Matrix<Type, 3, 1> Vector3;
+	typedef Matrix<Type, 3, 1> Vector3;
 
-    /**
-     * Standard constructor
-     *
-     * Initializes to identity
-     */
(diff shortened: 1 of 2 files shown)


jgoppert:
Looks good. Please add a unit test for it.

kd0aij:
should I also run "make format"? Will cause a lot of whitespace changes.

kd0aij:
I added a unit test in systemcmds/tests: https://github.com/PX4/Firmware/pull/5785 and tested.
But it's not clear how to add one here and build/run it.

jgoppert:
@kd0aij Here is how to run the testing/formatting:

```bash
git clone git@github.com:PX4/Matrix
mkdir build
cd build
cmake .. -DFORMAT=ON -DTESTING=ON
make format
make coverage
make test
ctest -V
```

jgoppert:
@kd0aij Please add a test here: https://github.com/PX4/Matrix/blob/master/test/attitude.cpp#L98

kd0aij:
@jgoppert "make coverage" fails with an incomprehensible error message, but the new test runs OK
"make format" changed every file in matrix/matrix but I only committed the files which I had edited.

jgoppert:
Closing in favor of #32.

Issue #30, opened 2016-08-22
Cholesky decomposition?


mkschreder:
Anyone thought of adding this in?

PR #34, opened 2016-12-10
Changes for hamilton quaternion convention.

### LINKED ISSUE #29: Support hamilton quaternion
Benefits:
- Same product convention as DCM. 
  - v_c = C_cb \* C_ba \* v_a  v_q = q_cb \* q_ba \* v_a
  - currently:  v_q = q_ba \* q_cb \* v_a (not very intuitive)
- More widely used (ROS) etc.

Drawback:
- ecl library already written assuming products are ordered the other way

@LorenzMeier @tumbili @bkueng @priseborough @mkschreder

jgoppert:
It is about time we got this switched over. Can we get the necessary changes in to EKF2 @priseborough . Closes #29.

The changes to the unit test should show you all that will need to be changed in EKF2.

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -4,11 +4,20 @@
  * All rotations and axis systems follow the right-hand rule.
  * The Hamilton quaternion product definition is used.
  *
- * In order to rotate a vector v by a righthand rotation defined by the quaternion q
+ * In order to rotate a vector in frame b (v_b) to frame n by a righthand
+ * rotation defined by the quaternion q_nb (from frame b to n)
  * one can use the following operation:
- * v_rotated = q^(-1) * [0;v] * q
- * where q^(-1) represents the inverse of the quaternion q.
- * The product z of two quaternions z = q1 * q2 represents an intrinsic rotation
+ * v_n = q_nb * [0;v_b] * q_nb^-1
+ *
+ * Just like DCM's: v_n = C_nb * v_b (vector rotation)
+ * M_n = C_nb * M_b * C_nb^(-1) (matrix rotation)
+ *
+ * or similarly
+ * v_b = q_nb^1 * [0;v_n] * q_nb
+ *
+ * where q_nb^(-1) represents the inverse of the quaternion q_nb =  q_bn
+ *
+ * The product z of two quaternions z = q2 * q1 represents an intrinsic rotation
  * in the order of first q1 followed by q2.
  * The first element of the quaternion
  * represents the real part, thus, a quaternion representing a zero-rotation
@@ -211,9 +220,9 @@ public:
         const Quaternion &p = *this;
         Quaternion r;
         r(0) = p(0) * q(0) - p(1) * q(1) - p(2) * q(2) - p(3) * q(3);
-        r(1) = p(0) * q(1) + p(1) * q(0) - p(2) * q(3) + p(3) * q(2);
-        r(2) = p(0) * q(2) + p(1) * q(3) + p(2) * q(0) - p(3) * q(1);
-        r(3) = p(0) * q(3) - p(1) * q(2) + p(2) * q(1) + p(3) * q(0);
+        r(1) = p(0) * q(1) + p(1) * q(0) + p(2) * q(3) - p(3) * q(2);
+        r(2) = p(0) * q(2) - p(1) * q(3) + p(2) * q(0) + p(3) * q(1);
+        r(3) = p(0) * q(3) + p(1) * q(2) - p(2) * q(1) + p(3) * q(0);
         return r;
     }
 
--- a/test/attitude.cpp
+++ b/test/attitude.cpp
@@ -170,13 +170,13 @@ int main()
     // quaternion derivative in frame 1
     Quatf q1(0, 1, 0, 0);
     Vector<float, 4> q1_dot1 = q1.derivative1(Vector3f(1, 2, 3));
-    float data_q_dot1_check[] = { -0.5f, 0.0f, 1.5f, -1.0f};
+    float data_q_dot1_check[] = { -0.5f, 0.0f, -1.5f, 1.0f};
     Vector<float, 4> q1_dot1_check(data_q_dot1_check);
     TEST(isEqual(q1_dot1, q1_dot1_check));
 
     // quaternion derivative in frame 2
     Vector<float, 4> q1_dot2 = q1.derivative2(Vector3f(1, 2, 3));
-    float data_q_dot2_check[] = { -0.5f, 0.0f, -1.5f, 1.0f};
+    float data_q_dot2_check[] = { -0.5f, 0.0f, 1.5f, -1.0f};
     Vector<float, 4> q1_dot2_check(data_q_dot2_check);
     TEST(isEqual(q1_dot2, q1_dot2_check));
 
@@ -296,8 +296,8 @@ int main()
 
     // conjugate
     Vector3f v1(1.5f, 2.2f, 3.2f);
-    TEST(isEqual(q.conjugate_inversed(v1), Dcmf(q)*v1));
-    TEST(isEqual(q.conjugate(v1), Dcmf(q).T()*v1));
+    TEST(isEqual(q.conjugate_inversed(v1), Dcmf(q).T()*v1));
+    TEST(isEqual(q.conjugate(v1), Dcmf(q)*v1));
 
     AxisAnglef aa_q_init(q);
     TEST(isEqual(aa_q_init, AxisAnglef(1.0f, 2.0f, 3.0f)));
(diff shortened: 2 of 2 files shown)


priseborough:
@jgoppert I want to keep the ecl EKF able to run as a standalone library without reliance on other PX4 libraries. I will have a look at the changes to conventions implemented in this PR and see how much work is involved to do this for all EKF operations.

jgoppert:
OK, my motivation would be to have ecl using matrix for math like the rest of the px4 code base to reduce the firmware size and make sure we have independent unit testing.

MaEtUgR:
The adaption to an existing quaternion convention is very welcome to me as I had problems with this librarie's quaternion convention in the past. The multiplication is the most important piece of the adaption and I'm really glad to see it change but I hope you are aware that other calculations like for example the dcm conversion also need to be compliant to the convention to have the perfectly predictable behavior.

I'm not sure if this is already the case and just wanted to mention. If you wish I would gladly take a look at all this soon.

jgoppert:
@MaEtUgR I believe we have covered all conversions in the unit tests. For instance if you look at the changes in this diff to unit testing you see the impact of the quaternion multplication changes. Basically, it just inverts the multiplication order.

MaEtUgR:
@priseborough as it was already mentioned this should only imply opposite order of all quaternion multiplications...

jgoppert:
Now also closes https://github.com/PX4/Matrix/issues/30.

matrix/Quaternion.hpp
```diff
@@ -280,9 +280,11 @@ class Quaternion : public Vector<Type, 4>
     void invert()
     {
         Quaternion &q = *this;
-        q(1) *= -1;
-        q(2) *= -1;
-        q(3) *= -1;
+        Type n = q.dot(q);
```
MaEtUgR: `q.dot(q)` gives you the norm squared but you want to devide by the norm here.
I would just use the norm method that we already have here `n = q.norm()` which would use this correct implementation: https://github.com/PX4/Matrix/blob/master/matrix/Vector.hpp#L73

matrix/Quaternion.hpp
```diff
@@ -294,10 +296,11 @@ class Quaternion : public Vector<Type, 4>
     {
         Quaternion &q = *this;
         Quaternion ret;
-        ret(0) = q(0);
-        ret(1) = -q(1);
-        ret(2) = -q(2);
-        ret(3) = -q(3);
+        Type n = q.dot(q);
```
MaEtUgR: same as last comment.
btw a test to identify this issue would be:
```
q = [.5 0 0 0]
q.inv() == [2 0 0 0]
```

matrix/Quaternion.hpp
jgoppert: Should have called the variable nsq to be clear. But for inverse believe it is conjugate over norm sq.

MaEtUgR: It is over the norm (without square) because the definition of the inverse is:
q * q^-1 = q^-1 * q = q_I    where the identity quaternion q_I = [1 0 0 0] has unit length.
You can infer from the quaternion multiplication that |q^-1| = 1/|q|.

The matlab implementation is over norm and you can also look up a definition somwhere else for example https://www.j3d.org/matrix_faq/matrfaq_latest.html for example says:
```
How do I calculate the inverse of a quaternion?
----------------------------------------------------
  This is equivalent to calculating the conjugate of the quaternion,
  if the quaternion is normalized (or a unit quaternion).
  In all other cases, the magnitude of the inverse is 1/|q|.
```

jgoppert: Here is the derivation with our quaternion product. You cannot assume unit length here. This derivation works for unit lenght and non-unit lenght quaternions: https://github.com/jgoppert/iekf_analysis/blob/master/Quaternion%20Inverse%20Check.ipynb

MaEtUgR: Thanks, I'll have a look at it asap.

MaEtUgR: I checked, you are right! sorry, my fault.

matrix/Matrix.hpp
```diff
@@ -290,8 +290,8 @@ class Matrix
 
     void print() const
     {
-        char buf[200];
-        write_string(buf, 200);
+        char buf[1000];
```
LorenzMeier: That's pretty evil on an embedded system. We need to find a better way than this.

MaEtUgR: It's only for debug prints but nevertheless I agree. What would be you proposition? Dynamic allocation?

dagar:
@jgoppert @priseborough we currently have matrix cloned twice in PX4 which is a bit confusing. Could we just use the version that comes in through ecl?

     db4374882bba122cb55370c5ea5efacbb8aa11c1 src/lib/ecl/matrix (v1.0.2-4-gdb43748)
     34fccdd6806fbdf7635a6365d7ef112f3bbaab9a src/lib/matrix (v1.0.2-5-g34fccdd)

jgoppert:
@dagar I'm usually bumping matrix a lot as I work on the estimators. I don't think we want to make IEKF and LPE dependent on ecl. Currently it seems ekf2 is grabbing the px4 version of matrix anywyas because it blows up when I pull in the hamilton quaternion patch. ecl should probably only have a copy of matrix lib for test purposes.

matrix/Matrix.hpp
MaEtUgR: Ok, but if dynamic then using the fields for the exact size. Otherwise we have two disadavantages.

matrix/Matrix.hpp
```diff
@@ -290,8 +290,8 @@ class Matrix
 
     void print() const
     {
-        char buf[1000];
-        write_string(buf, 1000);
+        char * buf = new char[15*N*M];
```
MaEtUgR: Do you really want to spend 15 characters for every number? Seems excessive to me... but maybe you have other applications.

matrix/Matrix.hpp
```diff
@@ -499,8 +499,8 @@ bool isEqual(const Matrix<Type, M, N> &x,
 
 
     if (!equal) {
-        char buf_x[100];
-        char buf_y[100];
+        char * buf_x = new char[100];
```
MaEtUgR: Why is it then not dynamic here? We already generate the char array on the heap for print.
The easiest solution could be something like:
```
printf("not equal:\n x:\n");
x.print();
printf("\ny:\n");
y.print();
```

jgoppert:
@MaEtUgR I think the static arrays is a good compromise with size based on the template. Let me know what you think. Can always go back to dynamic.

jgoppert:
I just compressed the commits so this is easier to review.

jgoppert:
It seems like coveralls stopped getting new builds. https://coveralls.io/github/PX4/Matrix

jgoppert:
This is cleaned up again to the minimal diff.

MaEtUgR:
Is there any concern with merging this or why is it still pending?

jgoppert:
@MaEtUgR We would need to get ekf2 converted over to this convention.

jgoppert:
@priseborough Can you help me get this into ekf2 so we can close it. You should just have to modify your derivation script for the new quaternion multiplication definition.

priseborough:
OK, let's set up a call and you can walk me through the changes. I do wish to keep the ecl EKF  portable and usable outside the PX4 framework.

jgoppert:
@priseborough great, I will contact you via skype.

MaEtUgR:
Switching multiplications should not be too much of a hassle, let me know if I can help anywhere to speed things up 👍

MaEtUgR:
@jgoppert @priseborough @LorenzMeier Can we take this in now please? I want to push the switch to the useful hamilton convention now. It's pending forever and if we are not switching soon the current way will also be used in a lot more interfaces (ask @julianoes ) and we will have the PX4 quaternion convention in future student books. Sorry for making this so urgent but otherwise nothing will ever happen. If noone writes back I'll merge and start converting multiplication to make submodule updates possible...

priseborough:
I will create a PR for the ecl library that uses the revised multiplication order and reference it to this PR.

priseborough:
I've been through the ecl and have a branch with the required changes https://github.com/PX4/ecl/tree/pr-ekfQuatMultOrder

I'm happy for this PR to be merged.

MaEtUgR:
Let's start testing and adapting in PRs for all the clients.

matrix/Quaternion.hpp
```diff
@@ -4,11 +4,20 @@
...
- * v_rotated = q^(-1) * [0;v] * q
- * where q^(-1) represents the inverse of the quaternion q.
- * The product z of two quaternions z = q1 * q2 represents an intrinsic rotation
+ * v_n = q_nb * [0;v_b] * q_nb^-1
+ *
+ * Just like DCM's: v_n = C_nb * v_b (vector rotation)
+ * M_n = C_nb * M_b * C_nb^(-1) (matrix rotation)
+ *
+ * or similarly
+ * v_b = q_nb^1 * [0;v_n] * q_nb
```
ndepal: A bit late to the party, but there is a `-` missing, should be `v_b = q_nb^-1 ...`.

MaEtUgR: I agree, there's another ^(-1) missing. Can you quickly review https://github.com/PX4/Matrix/pull/48 please?

MaEtUgR: Again late to the party, it's merged. Thanks for spotting 👍

PR #35, opened 2016-12-13
Fix for zero check on quat from dcm ctor.

jgoppert:
@MaEtUgR Thanks for spotting this. I'm not familiar with the pivot method in relation to dcm to quaternion conversion. I've corrected this to what I intended below. Let me know if this looks good to you.

### DIFF
--- a/matrix/Dcm.hpp
+++ b/matrix/Dcm.hpp
@@ -38,7 +38,7 @@ class AxisAngle;
  * described by this class.
  */
 template<typename Type>
-class Dcm : public Matrix<Type, 3, 3>
+class Dcm : public SquareMatrix<Type, 3>
 {
 public:
     virtual ~Dcm() {};
@@ -50,7 +50,7 @@ public:
      *
      * Initializes to identity
      */
-    Dcm() : Matrix<Type, 3, 3>()
+    Dcm() : SquareMatrix<Type, 3>()
     {
         (*this) = eye<Type, 3>();
     }
@@ -60,7 +60,7 @@ public:
      *
      * @param _data pointer to array
      */
-    Dcm(const Type *data_) : Matrix<Type, 3, 3>(data_)
+    Dcm(const Type *data_) : SquareMatrix<Type, 3>(data_)
     {
     }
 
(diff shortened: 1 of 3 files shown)


MaEtUgR:
@jgoppert I can implement the biggest pivot method.
It uses 4 different formulas for the same conversion and decides which formula to use such that the denominator is as far as possible from 0. This results in a globally numerically stable conversion.

jgoppert:
Sounds good, we may also want some case like this in case the user sends us a 3x3 zero matrix. This of course should be an exception but not sure how else to handle it for flight code.

MaEtUgR:
First try, hopefully I can fix the proplems shortly.

jgoppert:
@MaEtUgR Need to cover this case: https://coveralls.io/builds/9280785/source?filename=matrix%2FQuaternion.hpp#L113

MaEtUgR:
sure, on the way, and while covering it I found a potential error of mine.
tracking it now and will fix soon

MaEtUgR:
@jgoppert went quicker than I thought, I just had a typo in the test, I will try some random things to break it and then it's ready to merge

MaEtUgR:
@jgoppert did not break, ready to merge
corner cases as well as near corner cases and random cases worked fine

MaEtUgR:
and just to clarify why this change was indeed necessary for example if you take the example quaternion [0,1,0,0] and convert it to dcm and back the "standard" method went into division by zero and your fix just returned [1,0,0,0] what corresponds to a total different rotation

jgoppert:
Sorry I squashed so it didn't show that you authored most of that. Will make sure not to do that in the future.

MaEtUgR:
Ahem, sorry but that's NOT ok in my opinion, if you want I can create my own pull request and you can squash it to one commit but I still want credits for my changes. I worked for this...

LorenzMeier:
Let's apply those commits on top, which should be possible. Rewrite history if required.

MaEtUgR:
i'm not allowed to force push on master...

LorenzMeier:
I've done that for you to avoid additional churn. @jgoppert Sorry for interfering, its just really critical to get this right and given the short period I don't expect any issues with it.

MaEtUgR:
@LorenzMeier thanks, you are too kind
@jgoppert sorry for any inconvenience

jgoppert:
@LorenzMeier thanks for handling that. @MaEtUgR, understand,  by no means intended.

PR #41, opened 2017-03-17
Fixing misuse of the standard math library (performance audit)

pavel-kirienko:
Please refer there for context: https://github.com/PX4/Firmware/pull/6829#issuecomment-286920967.

The root of the problem is an improper use of the math library. The templated classes were implicitly making assumptions about the scalar type. For example, in many places the library used to make references to `::sinf` or `::sqrt`, essentially assuming that the scalar type is float or double, respectively.

Normally, this problem would be really easy to fix; consider the following two of the most obvious approaches:

* Replace all references to the C standard library with alternative references to the C++ standard library by adding `std::` where necessary. This approach is pretty straightorward, but lacks in two ways:
    * A lot of code modifications.
    * No proof against re-occurrence of the same problem in the future. This one could be solved by means of advanced static analysis though, but it would be slightly outside of the scope right now.
* Shadow the C math functions with C++ alternatives using the `using` directive. This approach requires minimal changes to the codebase and is absolutely resilient to re-occurrence of the problem in the future: in order to break things again, a developer would have to explicitly specify the desired scope of the referred definition, e.g. `::sin`, which people normally don't do.

N.B.: high integrity coding standards, such as MISRA C++ or HIC++, *consistently prohibit the use of the C standard library*. This specific case is a good anecdote in support of that recommendation.

The second solution from the above have been implemented here and tested against the unit tests (they all pass). Sadly, my attempt to test it with NuttX was unsuccessful due to the sad state of its C++ standard library. It turned out that instead of providing correct overloads in adherence to the C++ specification, it simply imports the C definitions into the `std` namespace. Consider `fabs`:

```c++
  using ::fabs;
```

That's it. This is what should have been there according to the standard:

```c++
float       fabs(float);
double      fabs(double);
long double fabs(long double);
```

The problem I described in the PX4 performance audit thread cannot be fixed right now without the loss of generality in the Matrix library. There are two solutions:

* Partially fix the NuttX C++ standard library. This is not something people do overnight.
* Implement the missing math functions manually in the Matrix library. Here's an idea how we can do that:

```c++
inline float       fabs(float x)       { return ::fabsf(x); }
inline double      fabs(double x)      { return ::fabs(x); }
inline long double fabs(long double x) { return ::fabsl(x); }
```

And so on. The list of missing functions is not large, I believe we only need about 10 or so.

### DIFF
--- a/matrix/stdlib_imports.hpp
+++ b/matrix/stdlib_imports.hpp
@@ -40,7 +40,7 @@ using std::frexp;
 using std::ldexp;
 using std::modf;
 
-#if __cplusplus >= 201103L
+#if (__cplusplus >= 201103L) && !defined(__PX4_NUTTX)
 
 using std::imaxabs;
 using std::imaxdiv;
--- a/matrix/stdlib_imports.hpp
+++ b/matrix/stdlib_imports.hpp
@@ -1,6 +1,9 @@
 /**
  * @file stdlib_imports.hpp
  *
+ * This file is needed to shadow the C standard library math functions with ones provided by the C++ standard library.
+ * This way we can guarantee that unwanted functions from the C library will never creep back in unexpectedly.
+ *
  * @author Pavel Kirienko <pavel.kirienko@zubax.com>
  */
 
@@ -36,18 +39,6 @@ using std::floor;
 using std::frexp;
 using std::ldexp;
 using std::modf;
-using std::fpclassify;
-using std::isfinite;
-using std::isinf;
-using std::isnan;
-using std::isnormal;
-using std::signbit;
-using std::isgreater;
-using std::isgreaterequal;
-using std::isless;
-using std::islessequal;
-using std::islessgreater;
-using std::isunordered;
 
 #if __cplusplus >= 201103L
 
@@ -84,6 +75,18 @@ using std::ilogb;
 using std::logb;
 using std::nextafter;
 using std::copysign;
+using std::fpclassify;
+using std::isfinite;
+using std::isinf;
+using std::isnan;
+using std::isnormal;
+using std::signbit;
+using std::isgreater;
+using std::isgreaterequal;
+using std::isless;
+using std::islessequal;
+using std::islessgreater;
+using std::isunordered;
 
 #endif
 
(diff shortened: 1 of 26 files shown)


pavel-kirienko:
I have applied the second workaround described above, now it works well with NuttX.

matrix/stdlib_imports.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+#pragma once
+
+#include <cmath>
+#include <cstdlib>
+#include <cinttypes>
+
+namespace matrix {
+
+#if defined(__PX4_NUTTX)
```
pavel-kirienko: Is there a way to detect NuttX in general instead of the PX4 flavor of it?

dagar:
Thanks for the thorough explanation Pavel. What about situations where the return is assigned to a float, but the input is a double. Wouldn't it be better to explicitly call the float version?

pavel-kirienko:
Not always. Use of double suggests that a higher precision is needed than what float can provide, so converting to float before the computation may have adverse effects on numerical precision.

matrix/stdlib_imports.hpp
jgoppert: @davids5

jgoppert:
Just waiting on feedback from @davids5 , then I think we are good to merge.

davids5:
@jgoppert LTGM

matrix/stdlib_imports.hpp
davids5: @pavel-kirienko

For the record no there is none. So far, a change to add __NUTTX__ to the generated config.h will not be accepted upstream. So this has to be set in the build by the flags, and we do that with __PX4_NUTTX.

dagar:
@pavel-kirienko do you know any tools that would help us work towards MISRA C++ or HIC++? I haven't looked recently, but the MISRA options I've seen so far have been quite expensive.

pavel-kirienko:
@dagar there is SciTools Understand, it's free for open source, but it tends to produce heaps of false positive. A bit unfriendly but usable.

PR #42, opened 2017-03-19
stdlib_imports don't limit to matrix namespace

dagar:
This allows the math fixes to be used in the rest of the firmware by including matrix/math.hpp. 

See https://github.com/PX4/Firmware/pull/6841

### DIFF
--- a/matrix/stdlib_imports.hpp
+++ b/matrix/stdlib_imports.hpp
@@ -13,8 +13,6 @@
 #include <cstdlib>
 #include <cinttypes>
 
-namespace matrix {
-
 #if defined(__PX4_NUTTX)
 /*
  * NuttX has no usable C++ math library, so we need to provide the needed definitions here manually.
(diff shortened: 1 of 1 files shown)


dagar:
No problem, it seemed questionable, but I created this to prompt the discussion for how to fix the remainder of the silent double conversions in Firmware. https://travis-ci.org/PX4/Firmware/jobs/212249160

By eliminate all references to the C library do you mean calls to fabs, fabsf, etc and `#include <math.h>? Should we use them from matrix going forward to get the Nuttx workaround? matrix::fabs, etc

pavel-kirienko:
> By eliminate all references to the C library do you mean calls to fabs, fabsf, etc and `#include <math.h>`?

Yes.

> Should we use them from matrix going forward to get the Nuttx workaround? matrix::fabs, etc

Excellent question. I don't know. Maybe the best way of handling this is just fixing the NuttX math library? At least in a minimal way? What would @davids5 say?

dagar:
LLVM libc++ for NuttX?

http://nuttx.org/doku.php?id=wiki:nshhowtos:llvm-stdcxx

davids5:
@dagar 

Yes - It will need to be tested on a branch tied to the current upstream (master_nxphlite).  
There was a lot of wide string routines, and local stubs added upstream to support C underpinnings. 
I expect the defconfig will have to be updated. We will need to look at the code size and performance.

dagar:
Clearly this needs more thought and it's not really a Matrix issue.

To summarize, there are still plenty of silent double conversions in PX4 Firmware.

PR #44, opened 2017-04-24
Add #define for mult type.

jgoppert:
This is a transition solution for ekf2 and iekf to get along until EKF2 is switched to the hamilton convention.


CMakeLists.txt
```diff
@@ -84,7 +89,7 @@ if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
 		)
 else()
 	add_compile_options(
-		-Wstrict-overflow=5
+		-Wstrict-overflow
```
jgoppert: This wouldn't compile on my local machine with =5

PR #48, opened 2017-09-25
Quaternion: Correcting Comments

MaEtUgR:
^(-1) went missing and some phrase was inprecise

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -2,20 +2,20 @@
  * @file Quaternion.hpp
  *
  * All rotations and axis systems follow the right-hand rule.
- * The Hamilton quaternion product definition is used.
+ * The Hamilton quaternion convention including its product definition is used.
  *
  * In order to rotate a vector in frame b (v_b) to frame n by a righthand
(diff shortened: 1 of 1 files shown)


matrix/Quaternion.hpp
```diff
@@ -2,20 +2,20 @@
...
  *
  * All rotations and axis systems follow the right-hand rule.
- * The Hamilton quaternion product definition is used.
+ * The Hamilton quaternion convention including its product definition is used.
  *
  * In order to rotate a vector in frame b (v_b) to frame n by a righthand
  * rotation defined by the quaternion q_nb (from frame b to n)
  * one can use the following operation:
- * v_n = q_nb * [0;v_b] * q_nb^-1
+ * v_n = q_nb * [0;v_b] * q_nb^(-1)
```
TSC21: I don't think it require the parenthesis

TSC21: Though those are comments so I don't see an issue

MaEtUgR: Not required at all, I just wanted to be consistent be it with or without.

PR #49, opened 2017-10-01
Vector: made unit inline

MaEtUgR:


LorenzMeier:
Could you please provide more context what the state was before and what this improves?

MaEtUgR:
Nevermind, after talking to @bkueng this seems completely useless. All those should be automatically optimized to be inline anyways.

PR #50, opened 2017-10-16
Matrix Print Buffer and Quaternion Fixes

MaEtUgR:
- Matrix `print()` buffer was usually to small and characters went missing. No memory corruption occured with a too small buffer because of the snprintf checks.
- I tried to improve the comments for Quaternion derivative and conjugate to allow people to quickly understand and hopefully also take advantage of the quaternion based functionality instead of always taking the detour to dcms.
- In my opinion I found a bug in the `rotate()` method sice we switched to the Hamilton convention (https://github.com/PX4/Matrix/pull/34). In the useful convention we commited to the quaternion multiplication order to rotate frames corresponds to the order that equivalent rotation matrices have and hence we need to multiply from the left to rotate a frame by a certain amount. (see section 2.6 on page 26 for reference http://www.iri.upc.edu/people/jsola/JoanSola/objectes/notes/kinematics.pdf). This bug went unnoticed because I could not find any module using the method and the test for this only cover tirival cases that do not explore non-commuting quaternions.

I know now is the moment to chatch the uncovered cases but I can't imagine an easy test opposed to just generating random rotations and use the result from other programs. If anyone has a better idea feel free to add it here.

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -276,11 +276,11 @@ public:
 
     /**
      * Computes the derivative of q_12 when
-     * rotated with angular velocity expressed in frame 2
+     * rotated with angular velocity expressed in frame 1
      * v_2 = q_12 * v_1 * q_12^-1
      * d/dt q_12 = 0.5 * q_12 * omega_12_2
      *
-     * @param w angular rate in frame 2
+     * @param w angular rate in frame 1 (typically body frame)
      */
     Matrix41 derivative1(const Matrix31 &w) const
     {
@@ -295,7 +295,7 @@ public:
      * v_2 = q_12 * v_1 * q_12^-1
      * d/dt q_12 = 0.5 * omega_12_1 * q_12
      *
-     * @param w angular rate in frame (typically reference frame)
+     * @param w angular rate in frame 2 (typically reference frame)
      */
     Matrix41 derivative2(const Matrix31 &w) const
     {
@@ -341,6 +341,14 @@ public:
         (*this) = (*this) * res;
     }
 
+    /**
+     * Rotates vector v_1 in frame 1 to vector r_2 in frame 2
+     * using the rotation quaternion q_12
+     * v_2 = q_12 * v_1 * q_12^-1
+     *
+     * @param vec vector to rotate in frame 1 (typically body frame)
+     * @return rotated vector in frame 2 (typically reference frame)
+     */
     Vector3f conjugate(const Vector3f &vec) {
         Quaternion q = *this;
         Quaternion v(0, vec(0), vec(1), vec(2));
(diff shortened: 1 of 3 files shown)


test/attitude.cpp
```diff
@@ -225,21 +225,22 @@ int main()
...
     rot(1) = rot(2) = 0.0f;
     qI.rotate(rot);
     q_true = Quatf(cos(0.0f), sin(0.0f), 0.0f, 0.0f);
-    TEST(fabs(qI(0) - q_true(0)) < eps);
-    TEST(fabs(qI(1) - q_true(1)) < eps);
-    TEST(fabs(qI(2) - q_true(2)) < eps);
-    TEST(fabs(qI(3) - q_true(3)) < eps);
+    TEST(isEqual(qI, q_true));
+
+    // rotate quaternion (random non-commutating rotation)
```
MaEtUgR: I added the missing test which catches non-trivial cases here.

MaEtUgR:
- switched `rotate()` to the AxisAngle call like the `TODO` comment suggested
- added the test for non-trivial cases with result from Matlab (any double checks appreciated)
it immediately succedded because of https://github.com/PX4/Matrix/pull/50/commits/aece6d993bc49965c5cca116e7950408cb11251a

MaEtUgR:
As a sanity check to be sure I didn't overlook something:
1. before the hamilton convention switch my additional test succeeds: https://github.com/PX4/Matrix/pull/52
1. after the hamilton convention switch without this pr my additional test is broken (because calculation is wrong in the current master): https://github.com/PX4/Matrix/pull/51
1. with this pr it's fixed again 👍

LorenzMeier:
@jgoppert Could you review?

matrix/Quaternion.hpp
```diff
@@ -330,24 +330,38 @@ class Quaternion : public Vector<Type, 4>
...
-        (*this) = (*this) * res;
+        Quaternion res(vec);
+        (*this) = res * (*this);
     }
 
+    /**
+     * Rotates vector v_1 in frame 1 to vector r_2 in frame 2
+     * using the rotation quaternion q_12
+     * v_2 = q_12 * v_1 * q_12^-1
+     *
```
Stifael: what is exactly the convention for the naming: q_12 vs q_21

matrix/Quaternion.hpp
```diff
@@ -330,24 +330,38 @@ class Quaternion : public Vector<Type, 4>
...
     Vector3f conjugate(const Vector3f &vec) {
         Quaternion q = *this;
         Quaternion v(0, vec(0), vec(1), vec(2));
         Quaternion res = q*v*q.inversed();
         return Vector3f(res(1), res(2), res(3));
     }
 
+    /**
+     * Rotates vector v_2 in frame 2 to vector r_1 in frame 1
+     * using the rotation quaternion q_12
```
Stifael: v_1

Stifael:
and thanks for the paper

matrix/Quaternion.hpp
MaEtUgR: You are totally right, the indexing convention is widespread the opposite way (also in the paper I'm referncing in the original post). I'll change the comments.

MaEtUgR:
@Stifael I corrected the typos and index inconsistencies you found in the comments, thanks 👍

PR #54, opened 2017-11-05
Normalization with Zero Check, fixed stub test again

MaEtUgR:
Original idea:
I came across normalization with zero check again and again and again in the code. Why not make it part of the library? It's not default, you still get `[NaN,NaN,NaN]` when executing `[0,0,0].unit();` but in case you already know you want a zero vector to stay zero, let's offer it.

While writing the tests for the new method I had a closer look at the vector.cpp tests and found:
- `make check` didn't run locally on GCC 5.4.0 for me because of loop unrolling signed overflow warning together with -Werror resulted in compile errors. It seems that CI does less optimization for the tests...
- `normalize()` is **another stub test** which only gets called but not evaluated
- no structure and comments
- the `operator*` dot product check reinitialized all data unnecessarily

So I fixed all the problems I found and wrote my test. It should be all fine now.

### DIFF
--- a/matrix/Vector.hpp
+++ b/matrix/Vector.hpp
@@ -83,6 +83,14 @@ public:
         return (*this) / norm();
     }
 
+    Vector unit_or_zero(const Type eps = 1e-5f) {
+        const Type n = norm();
+        if (n > eps) {
+            return (*this) / n;
+        }
+        return Vector();
+    }
+
     inline Vector normalized() const {
         return unit();
     }
--- a/test/vector.cpp
+++ b/test/vector.cpp
@@ -6,25 +6,34 @@ using namespace matrix;
 
 int main()
 {
+    // test data
     float data1[] = {1,2,3,4,5};
     float data2[] = {6,7,8,9,10};
     Vector<float, 5> v1(data1);
-    TEST(isEqualF(v1.norm(), 7.416198487095663f));
-    TEST(isEqualF(v1.norm(), v1.length()));
     Vector<float, 5> v2(data2);
-    TEST(isEqualF(v1.dot(v2), 130.0f));
-    v2.normalize();
+
+    // copy constructor
     Vector<float, 5> v3(v2);
     TEST(isEqual(v2, v3));
+
+    // norm, dot product
+    TEST(isEqualF(v1.norm(), 7.416198487095663f));
+    TEST(isEqualF(v1.norm(), v1.length()));
+    TEST(isEqualF(v1.dot(v2), 130.0f));
+    TEST(isEqualF(v1.dot(v2), v1 * v2));
+
+    // unit, unit_zero, normalize
+    TEST(isEqualF(v2.unit().norm(), 1.f));
+    TEST(isEqualF(v2.unit_or_zero().norm(), 1.f));
+    TEST(isEqualF(Vector<float, 5>().unit_or_zero().norm(), 0.f));
+    v2.normalize();
+    TEST(isEqualF(v2.norm(), 1.f));
+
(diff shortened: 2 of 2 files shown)


matrix/Vector.hpp
```diff
@@ -83,6 +83,14 @@ class Vector : public Matrix<Type, M, 1>
         return (*this) / norm();
     }
 
+    Vector unit_or_zero(const Type eps = 1e-5f) {
+        const Type n = norm();
+        if (n > eps) {
+            return (*this) / norm();
```
bkueng: `return (*this) / n;`

test/attitude.cpp
```diff
@@ -107,8 +107,9 @@ int main()
     double rad2deg = 180.0 / M_PI;
 
     // euler dcm round trip check
-    for (int roll = -90; roll <= 90; roll += 90) {
-        for (int pitch = -90; pitch <= 90; pitch += 90) {
+    // volatile to prevent loop unrolling and signed overflow warning as error
+    for (volatile int roll = -90; roll <= 90; roll += 90) {
+        for (volatile int pitch = -90; pitch <= 90; pitch += 90) {
```
bkueng: What is the exact error that you get? Is it just a GCC anomaly? `make check` runs through fine for me with GCC 7.2.1

matrix/Vector.hpp
MaEtUgR: This was my intention, don't know how it got lost and unnoticed... fixed now.

test/attitude.cpp
MaEtUgR:
```
maetugr @ ubuntu ~/Firmware/src/lib/matrix (normalization-with-zero-check) ▶ make check
Scanning dependencies of target attitude
[  2%] Building CXX object test/CMakeFiles/attitude.dir/attitude.cpp.o
/home/maetugr/Firmware/src/lib/matrix/test/attitude.cpp: In function ‘int main()’:
/home/maetugr/Firmware/src/lib/matrix/test/attitude.cpp:6:5: error: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Werror=strict-overflow]
 int main()
     ^
/home/maetugr/Firmware/src/lib/matrix/test/attitude.cpp:6:5: error: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Werror=strict-overflow]
cc1plus: all warnings being treated as errors
test/CMakeFiles/attitude.dir/build.make:62: recipe for target 'test/CMakeFiles/attitude.dir/attitude.cpp.o' failed
make[3]: *** [test/CMakeFiles/attitude.dir/attitude.cpp.o] Error 1
CMakeFiles/Makefile2:822: recipe for target 'test/CMakeFiles/attitude.dir/all' failed
make[2]: *** [test/CMakeFiles/attitude.dir/all] Error 2
CMakeFiles/Makefile2:74: recipe for target 'CMakeFiles/check.dir/rule' failed
make[1]: *** [CMakeFiles/check.dir/rule] Error 2
Makefile:151: recipe for target 'check' failed
make: *** [check] Error 2
```
Let's put this problem in a separate issue/pr then because it has nothing to do with the normalization and if it works without error on other systems/newer GCC the solution might not be right one.

MaEtUgR: I dropped this commit.

bkueng: Ok, it's just that GCC cannot determine whether an int overflow can occur or not. But I don't see that it can, so adding the volatile with a clear comment is fine for me.

PR #55, opened 2017-11-13
Quaternion: Shortest rotation between two vectors

MaEtUgR:
There are multiple application scenarios but I need this functionality for the quaternion controller: https://github.com/PX4/Firmware/pull/8003

And because of the reusability and the calculation itself suffering tedious corner cases I added it to the library with unit tests to ensure proper results.

A simple explanation for the corner cases: The input vectors are parallel but oposite and therefore require a 180 degree rotation but the cross product to get the perpendicular rotation axis is zero. For more discussion see: https://stackoverflow.com/questions/1171849/finding-quaternion-representing-the-rotation-from-one-vector-to-another .

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -186,6 +186,47 @@ public:
         }
     }
 
+    /**
+     * Quaternion from two vectors
+     * Generates shortest rotation from source to destination vector
+     * Default destination if not specified is [0,0,1]
+     *
+     * @param dst destination vector (no need to normalize)
+     * @param src source vector (no need to normalize)
+     * @param eps epsilon threshold which decides if a value is considered zero
+     */
+    Quaternion(const Vector3<Type> &src, const Vector<Type, 3> dst = Vector3<Type>(0, 0, 1), const Type eps = 1e-5f) :
+        Vector<Type, 4>()
+    {
+        Quaternion &q = *this;
+        Vector3<Type> cr = src.cross(dst);
+        float dt = src.dot(dst);
+        if (cr.norm() < eps && dt < 0) {
+            cr = src.abs();
+            if (cr(0) < cr(1)) {
+                if (cr(0) < cr(2)) {
+                    cr = Vector3<Type>(1, 0, 0);
+                } else {
+                    cr = Vector3<Type>(0, 0, 1);
+                }
+            } else {
+                if (cr(1) < cr(2)) {
+                    cr = Vector3<Type>(0, 1, 0);
+                } else {
+                    cr = Vector3<Type>(0, 0, 1);
+                }
+            }
+            q(0) = Type(0);
+            cr = src.cross(cr);
+        } else {
+            q(0) = src.dot(dst) + sqrt(src.norm_squared() * dst.norm_squared());
+        }
+        q(1) = cr(0);
+        q(2) = cr(1);
+        q(3) = cr(2);
+        q.normalize();
+    }
+
 
     /**
      * Constructor from quaternion values
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -189,18 +189,19 @@ public:
     /**
      * Quaternion from two vectors
      * Generates shortest rotation from source to destination vector
-     * Default destination if not specified is [0,0,1]
      *
      * @param dst destination vector (no need to normalize)
      * @param src source vector (no need to normalize)
      * @param eps epsilon threshold which decides if a value is considered zero
      */
-    Quaternion(const Vector3<Type> &src, const Vector<Type, 3> dst = Vector3<Type>(0, 0, 1), const Type eps = 1e-5f) :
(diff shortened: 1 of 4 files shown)


MaEtUgR:
Fixed coverage, there was a typo like error which resulted in not covering the case I designed the test for. I made the tests even simpler such that this cannot happen anymore.

matrix/Quaternion.hpp
```diff
@@ -186,6 +186,47 @@ class Quaternion : public Vector<Type, 4>
...
+    /**
+     * Quaternion from two vectors
+     * Generates shortest rotation from source to destination vector
+     * Default destination if not specified is [0,0,1]
+     *
+     * @param dst destination vector (no need to normalize)
+     * @param src source vector (no need to normalize)
+     * @param eps epsilon threshold which decides if a value is considered zero
+     */
+    Quaternion(const Vector3<Type> &src, const Vector<Type, 3> dst = Vector3<Type>(0, 0, 1), const Type eps = 1e-5f) :
```
Stifael: - Vector3<type> dst = Vector3<Type>(0,0,1)
- const Type eps = (Type)1e-5;

matrix/Quaternion.hpp
```diff
@@ -186,6 +186,47 @@ class Quaternion : public Vector<Type, 4>
...
+                if (cr(1) < cr(2)) {
+                    cr = Vector3<Type>(0, 1, 0);
+                } else {
+                    cr = Vector3<Type>(0, 0, 1);
+                }
+            }
+            q(0) = Type(0);
+            cr = src.cross(cr);
+        } else {
+            q(0) = src.dot(dst) + sqrt(src.norm_squared() * dst.norm_squared());
```
Stifael: this function is definitely not obvious.
maybe write something like "Half-Way Quaternion Solution".

matrix/Quaternion.hpp
```diff
@@ -186,6 +186,47 @@ class Quaternion : public Vector<Type, 4>
...
+                }
+            } else {
+                if (cr(1) < cr(2)) {
+                    cr = Vector3<Type>(0, 1, 0);
+                } else {
+                    cr = Vector3<Type>(0, 0, 1);
+                }
+            }
+            q(0) = Type(0);
+            cr = src.cross(cr);
```
Stifael: I think I get it what you do here with the cr manipulation, but I cannot really verify that it is correct or not? 
Do you have a link to an example?

MaEtUgR: There are a bulk of examples here: https://stackoverflow.com/questions/1171849/finding-quaternion-representing-the-rotation-from-one-vector-to-another

Also if you look at the unit tests you see in the comments which one is going through which case. Plus I can share the MATLAB code I used for intial prototyping of the functionality.

matrix/Quaternion.hpp
MaEtUgR: There was `Vector<Type, 3>` which I switched to `Vector3<Type>` now, that's what you meant right?
`const Type eps = (Type)1e-5` is enough, constants are casted auatomatically to the correct type on compile time.

matrix/Quaternion.hpp
```diff
@@ -186,6 +186,47 @@ class Quaternion : public Vector<Type, 4>
...
+     * @param src source vector (no need to normalize)
+     * @param eps epsilon threshold which decides if a value is considered zero
+     */
+    Quaternion(const Vector3<Type> &src, const Vector<Type, 3> dst = Vector3<Type>(0, 0, 1), const Type eps = 1e-5f) :
+        Vector<Type, 4>()
+    {
+        Quaternion &q = *this;
+        Vector3<Type> cr = src.cross(dst);
+        float dt = src.dot(dst);
+        if (cr.norm() < eps && dt < 0) {
```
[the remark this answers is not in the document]
MaEtUgR: You're totally right, I added comments for that.

matrix/Quaternion.hpp
MaEtUgR: comment added

MaEtUgR:
@Stifael Please have another look.

matrix/Quaternion.hpp
MaEtUgR: You're right. Second statement is wrong... I put the explicit conversion now.

matrix/Vector.hpp
```diff
@@ -71,6 +71,11 @@ class Vector : public Matrix<Type, M, 1>
         return Type(sqrt(a.dot(a)));
     }
 
+    Type norm_squared() const {
+        const Vector &a(*this);
+        return Type(a.dot(a));
```
Stifael: Is it actually necessary to cast it?

MaEtUgR: No, you're right, I forgot to get rid of it. It was only necessary for sqrt() which can return double...

MaEtUgR:
I added the fix (took out the cast) to the latest commit.

PR #40, opened 2017-02-24
Add Quaternion copyTo

dagar:
The main purpose of this PR is to add Quaternion copyTo (da47d65)

- add Quaternion copyTo to make it easier to copy a Quatf back into a uORB quat array - https://github.com/PX4/Firmware/compare/master...dagar:copyTo#diff-18e8f57d61a55777f8e25da991013704L733

 - updated astyle to 2.06 (same as PX4 and DriverFramework) and fixed formatting
 - add `make check` target that builds, runs testing, and checks formatting
 - added travis-ci builds for clang and OSX xcode
 - made a bunch of trivial changes in line with cppcoreguidelines
 - added options to run tests under address sanitizer and undefined behaviour sanitizer, but didn't find anything interesting

PR #56, opened 2017-11-15
Vector3: copyTo for reuse

MaEtUgR:
Similar to https://github.com/PX4/Matrix/pull/40 I added a Vector3.copyTo because there exist a lot of three line vector copy code blocks to fill uorb topics and this can be solved with this pr.

### DIFF
--- a/matrix/Dcm.hpp
+++ b/matrix/Dcm.hpp
@@ -153,6 +153,12 @@ public:
         dcm = Quaternion<Type>(aa);
     }
 
+    /**
+     * Vee operator get the vector of a skew symmetric matrix
+     * inverse operation: Vector3.hat()
+     *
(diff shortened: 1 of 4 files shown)


dagar:
Looks good, although we could push this to the base Vector or even Matrix (row major array copy?).

MaEtUgR:
@dagar sounds useful... I thought of that. Wasn't sure why you made it explicitly for quaternions and thought it's because there's actual array length compiler checks but taht's probably also achievable with the vector/matrix size.

dagar:
It's an oversight. Most of the PX4 usage was with Quaternions. The size is a template argument.

MaEtUgR:
@dagar sure, I'm preparing a pr.

MaEtUgR:
ready #57

PR #57, opened 2017-11-15
CopyTo functionality for any Matrix

MaEtUgR:
Based on the suggestion/discussion here: https://github.com/PX4/Matrix/pull/56#issuecomment-344663324 I add a copyTo method for generally any Matrix including vectors and quaternions for which is supposedly used the most.

- Because copying memory only allows for row-major order I wrote a method to also support **column-major** order.
- I include all the unit tests of course which deserve their **own test script** for clarity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ Makefile
 test/attitude
 test/cmake_install.cmake
 test/CMakeFiles/
+test/copyto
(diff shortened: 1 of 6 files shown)


dagar:
Looks good to me. I'm not sure when column major would actually be used, but sure why not?

PR #58, opened 2017-11-16
Matrix: added copyToRaw method to allow copying to a pointer

MaEtUgR:
because most uORB messages still contain all components of a vector one by one after each other

I discussed with @bkueng that the messages should stay like they are short-term for compatibility with flight review.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -87,6 +87,11 @@ public:
         memcpy(dst, _data, sizeof(dst));
     }
 
+    void copyToRaw(Type *dst) const
+    {
+        memcpy(dst, _data, sizeof(_data));
+    }
+
     void copyToColumnMajor(Type (&dst)[M*N]) const
     {
         const Matrix<Type, M, N> &self = *this;
--- a/test/copyto.cpp
+++ b/test/copyto.cpp
@@ -38,6 +38,13 @@ int main()
         TEST(fabs(array_A[i] - array_row[i]) < eps);
     }
 
+    // Matrix copyTo with a pointer
+    A.copyToRaw(static_cast <float *> (array_A));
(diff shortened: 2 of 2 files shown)


dagar:
You want to use this to copy into a uORB message starting at an arbitrary point? We can probably get away with it, but I'm not sure if it's worth it. I'd rather work towards something more restrictive and memory safe.
 
I think we should either update the messages themselves (+ what that breaks), add helpers on the PX4/uORB side, or just do it per element for now.

MaEtUgR:
@dagar That was exactly my thinking but:
- The element wise stuff is there for too long, I wanted to change that 1.5 years ago and it's still exactly like it was so I want to make it more expressive and convenient.
- I added the memory safe method already (copyTo) and started changing the messages and then @bkueng told me it would get a nightmare for flight review because it needs to support old logs and new logs and if the messages change one by one it gets unmaintainable...

I and probably also @bkueng would prefer the memory safe solution but if it takes so much effort we have to probably push back on that and do an organized switch of all messages to arrays at once.

bkueng:
> + what that breaks

That's what I want to avoid: creating & maintaining compatibility layers for such small changes for flight review and replay. I support the change, having type-safety is definitely desirable. But we should do it within a larger refactoring, for example when we split up messages or similar.

dagar:
Fair enough, but I'm also suggesting waiting rather than introducing something unsafe for a small (or likely negligible) performance improvement.

From my perspective I've seen quite a few people with weaker computing backgrounds try to jump and modify the controllers. Any additional holes we provide for mistakes will be found and used.

MaEtUgR:
@dagar It's less about performance but more about code reuse/readability in my opinion. For performance alone I agree it would not be worth it.

I suggest to take this method out again on the day that we are allowed to switch the messages containing vectors to represented them by arrays with a size that can be checked typesafe.

PR #59, opened 2017-12-04
LICENSE minor changes to template

dagar:
These minor changes allow github to recognize the license and display the tag.

I believe only the first line change is actually needed.


dagar:
I wasn't sure if the year range actually means anything.

PR #60, opened 2018-03-19
Matrix add == and != operators

dagar:


matrix/Matrix.hpp
```diff
@@ -288,6 +288,29 @@ class Matrix
         *this = (*this) - scalar;
     }
 
+    bool operator==(const Matrix<Type, M, N> &other) const
+    {
+        const Matrix<Type, M, N> &self = *this;
+
+        // TODO: set this based on Type
+        static constexpr float eps = 1e-4f;
```
MaEtUgR: Why not make `eps` an optional parameter with default? So you can customize the expected accuracy as client?
That's already done in several places:
- https://github.com/PX4/Matrix/blob/f835d390171ae61de977d00cdaf5ed16e5361e4e/matrix/Vector.hpp#L91
- https://github.com/PX4/Matrix/blob/41ad2bdea52771a8f486a0be6098e6e45786bdcd/matrix/Matrix.hpp#L505
- https://github.com/PX4/Matrix/blob/f835d390171ae61de977d00cdaf5ed16e5361e4e/matrix/Quaternion.hpp#L197

dagar: I wasn't sure of a good way to do that via operator overloading. You can't pass additional arguments. With templates you can use enable_if to have different things for different types, but it gets ugly quickly.

dagar:
@MaEtUgR good to merge? We can revisit unifying equality with the tests later. This is needed to fully drop the old mathlib matrix in Firmware.

matrix/Matrix.hpp
MaEtUgR: Then lets add that option to isEqual via argument when we merge them or find an even better solution based on types.

MaEtUgR:
Btw why is coverage still 100%? You didn't add any test and it's new functionality...

dagar:
Thanks, sounds good to me. We need to figure out what we're even doing with the test quite and where the print should live.

Not sure what's up with the coverage tool.

image

MaEtUgR:
Also just found out it's ignoring coverage...

dagar:
The next pass can cleanup testing (possibly refactor isEqual) and figure out how to get real coverage numbers. Currently the test seems to be telling us that 100% of the code used in testing is covered.

jgoppert:
Testing was working and was at 100% last time I checked, let's not merge anything that isn't tested going forward. If coverage is broken we should fix that first before merging.

jgoppert:
You can see it in LCOV, it missed the functions.

screenshot from 2018-03-27 16-59-52

```
cmake -DCMAKE_BUILD_TYPE=Coverage -DTESTING=ON ..
make coverage_html
```

PR #63, opened 2018-03-27
Pr test equality

dagar:


jgoppert:
@dagar Oops we duplicated.

jgoppert:
I merged this with #64, so closing.

Issue #62, opened 2018-03-27
Assert Function Coverage 100%


jgoppert:
We need a way to tell coveralls to assert that function coverage is 100%. For template functions that are not used they are thrown away in the source and it reports 100% coverage when it doesn't even call the functions.

jgoppert:
Closed via #64. Needed to instantiate all templates to get accurate line coverage.

PR #64, opened 2018-03-27
Fix coverage and bug in matrix equal test.

jgoppert:
Towards closing #62. Still need to get coveralls to report function coverage. Not sure why function coverage isn't 100% now. It looks like I got everything in lcov.


matrix/Matrix.hpp
```diff
@@ -297,7 +297,7 @@ class Matrix
 
         for (size_t i = 0; i < M; i++) {
             for (size_t j = 0; j < N; j++) {
-                if (fabs(self(i, j) - other(i, j) > eps) ) {
+                if (fabs(self(i, j) - other(i, j)) > eps) {
```
jgoppert: @dagar :-P

test/attitude.cpp
```diff
@@ -130,18 +147,18 @@ int main()
     double rad2deg = 180.0 / M_PI;
 
     // euler dcm round trip check
-    for (int roll = -90; roll <= 90; roll += 90) {
-        for (int pitch = -90; pitch <= 90; pitch += 90) {
-            for (int yaw = -179; yaw <= 180; yaw += 90) {
+    for (double roll = -90; roll <= 90; roll += 90) {
+        for (double pitch = -90; pitch <= 90; pitch += 90) {
+            for (double yaw = -179; yaw <= 180; yaw += 90) {
                 // note if theta = pi/2, then roll is set to zero
```
jgoppert: using doubles here fixed a compile if you try to compile the tests without coverage flags on, not sure if we should add that to travis or not, not really a big deal

PR #65, opened 2018-03-28
Fix README/cmake format.

jgoppert:
I think it would be good to be consistent about using spaces instead of indent in the code and cmake files.


dagar:
I don't actually have a preference either way as long as it's consistent. I've been meaning to give cmake-lint a try at some point. https://github.com/richq/cmake-lint

PR #66, opened 2018-03-31
Quaternion mark const helpers const

dagar:
This is necessary if you want to call these methods on const Quaternions (depending on compiler warnings).


MaEtUgR:
Makes totally sense, also adds to in code documentation.

PR #67, opened 2018-04-19
Add 'getMatrix' method into 'Quaternion' class

xahon:
Added this very useful method (http://irrlicht.sourceforge.net/docu/quaternion_8h_source.html#l00349)

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -521,6 +521,41 @@ public:
         return Dcm<Type>(*this);
     }
 
+
+    SquareMatrix<Type, 4> get_matrix(const Vector<Type, 3> center) {
+      Type x = this(0), y = this(1), z = this(2), w = this(3);
+
+      /*
+       * Source (http://irrlicht.sourceforge.net/docu/quaternion_8h_source.html#l00349)
+       *
+       * [  1 - 2yy - 2zz     2xy + 2zw       2xz - 2yw       0  ]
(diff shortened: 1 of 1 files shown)


bresch:
Please note that this library uses the Hamiltonian convention as described here:
https://dev.px4.io/en/contribute/notation.html
px4_quat

MaEtUgR:
@xahon 
As @bresch correctly pointed out the library uses Hamiltonian convention and needs to be consistent.
Also you don't need to reimplement the quaternion to rotation matrix calculation but could rather use the existing method similar to this untested example:
```
/*
 * Useful Description like:
 * Build transformation matrix from quaternion and position coordinate vector to ...
 *
 * [ R(q)  center ]
 * [  0      1    ]
 *
 * Source (http://irrlicht.sourceforge.net/docu/quaternion_8h_source.html#l00349)
 */
SquareMatrix<Type, 4> transformationMatrix(const Vector3 center)
{
    const Quaternion &q = *this;
    SquareMatrix<Type, 4> T; // initialize result matrix with all elements zero

    // place the 3x3 rotation matrix built from q in the top left corner of T
    T.set<3, 3>(Dcmf(q), 0, 0);
    // place 3x1 vector center in the top right corner of T
    T.set<3, 1>(center, 3, 0);
    // place a 1 casted to the used type of the template in the lower right corner of T
    T(3, 3) = Type(1);

    return T;
}
```

Feel free to reopen if you follow up further on adding this functionality.

PR #68, opened 2018-06-09
helper_functions add wrap_2pi

dagar:


matrix/helper_functions.hpp
```diff
@@ -29,16 +29,31 @@ Type wrap_pi(Type x)
...
 
+template<typename Type>
+Type wrap_2pi(Type x)
+{
+    if (!is_finite(x)) {
+        return x;
+    }
+
+    while (x >= Type(2 * M_PI)) {
+        x -= Type(2.0 * M_PI);
```
LorenzMeier: This can still take a considerable amount of CPU cycles - should this be bounded? I would feel better if it had a limit of 100 iterations and would then return NaN or a similar strategy.

dagar: Couldn't hurt.

PR #71, opened 2018-06-27
Improve wrapping helper functions including tests

MaEtUgR:
~Using a loop to recreate a division is not a good solution and the additional limit for the cycle count doesn't make it better. Instead there exists efficient functionality in cmath to calculate
the remainder of a floating point devision: fmod.~

~Probably we can make it even more efficient by inlining which I suppose the compiler already does automatically. And adding a check if the value is already in range but I'm not sure if that really safes any compute.~

**EDIT:** I learned my lesson about `fmod` being among the least efficient math library methods I've seen both in compute and memory.

Instead this pr brings:
- generalized wrapping for any range
- more wrapping tests
- a fix for the `isEqual` method that's only used in testing but was duplicating the `print()` method while still having a too small static buffer size for the output

### DIFF
--- a/matrix/helper_functions.hpp
+++ b/matrix/helper_functions.hpp
@@ -20,64 +20,57 @@ bool is_finite(Type x) {
 #endif
 }
 
+/**
+ * Wrap value to stay in range [low, high)
+ *
+ * @param x input possibly outside of the range
+ * @param low lower limit of the allowed range
+ * @param high upper limit of the allowed range
+ * @return wrapped value inside the range
+ */
+template<typename Type>
+Type wrap(Type x, Type low, Type high) {
+    // already in range
+    if (low < x && x < high) {
+        return x;
+    }
+
+    // close to range
+    Type range = high - low;
+    if ((high <= x) && (x < high + (range*100))) {
+        while (high <= x) {
+            x -= range;
+        }
+        return x;
+    }
+
+    if ((low - (range*100) <= x) && (x < low)) {
+        while (x < low) {
+            x += range;
+        }
+        return x;
+    }
+
+    // very far from the range -> something went terribly wrong
+    return NAN;
+}
+
+/**
+ * Wrap value in range [-π, π)
+ */
 template<typename Type>
 Type wrap_pi(Type x)
 {
-    if (!is_finite(x)) {
-        return x;
-    }
-
-    int c = 0;
-
-    while (x >= Type(M_PI)) {
-        x -= Type(2 * M_PI);
-
-        if (c++ > 100) {
-            return INFINITY;
-        }
-    }
-
-    c = 0;
-
-    while (x < Type(-M_PI)) {
-        x += Type(2 * M_PI);
-
-        if (c++ > 100) {
-            return INFINITY;
-        }
-    }
-
-    return x;
+    return wrap(x, Type(-M_PI), Type(M_PI));
 }
 
+/**
+ * Wrap value in range [0, 2π)
+ */
 template<typename Type>
 Type wrap_2pi(Type x)
 {
-    if (!is_finite(x)) {
-        return x;
-    }
-
-    int c = 0;
-
-    while (x >= Type(2 * M_PI)) {
-        x -= Type(2 * M_PI);
-
-        if (c++ > 100) {
-            return INFINITY;
-        }
-    }
-
-    c = 0;
-
-    while (x < Type(0)) {
-        x += Type(2 * M_PI);
-
-        if (c++ > 100) {
-            return INFINITY;
-        }
-    }
-
-    return x;
+    return wrap(x, Type(0), Type(M_TWOPI));
 }
 
 }
--- a/matrix/helper_functions.hpp
+++ b/matrix/helper_functions.hpp
@@ -26,7 +26,7 @@ bool is_finite(Type x) {
  * @param x input possibly outside of the range
  * @param low lower limit of the allowed range
  * @param high upper limit of the allowed range
- * @return wrapped value inside the range
+ * @return wrapped value inside the range, or NAN if value is too far away from range.
  */
 template<typename Type>
 Type wrap(Type x, Type low, Type high) {
--- a/matrix/stdlib_imports.hpp
+++ b/matrix/stdlib_imports.hpp
@@ -13,6 +13,13 @@
 #include <cstdlib>
 #include <inttypes.h>
 
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+#ifndef M_TWOPI
+#define M_TWOPI (M_PI * 2.0)
+#endif
+
 namespace matrix {
 
 #if !defined(FLT_EPSILON)
--- a/test/helper.cpp
+++ b/test/helper.cpp
@@ -5,22 +5,35 @@ using namespace matrix;
 
 int main()
 {
-    TEST(fabs(wrap_pi(4.0) - (4.0 - 2*M_PI)) < FLT_EPSILON);
-    TEST(fabs(wrap_pi(-4.0) - (-4.0 + 2*M_PI)) < FLT_EPSILON);
-    TEST(fabs(wrap_pi(3.0) - (3.0)) < FLT_EPSILON);
-    TEST(!is_finite(wrap_pi(1000.0f)));
-    TEST(!is_finite(wrap_pi(-1000.0f)));
-    wrap_pi(NAN);
+    // general wraps
+    TEST(fabs(wrap(4.0, 0.0, 10.0) - 4.0) < FLT_EPSILON);
+    TEST(fabs(wrap(4.0, 0.0, 1.0)) < FLT_EPSILON);
+    TEST(fabs(wrap(-4.0, 0.0, 10.0) - 6.0) < FLT_EPSILON);
+    TEST(fabs(wrap(-18.0, 0.0, 10.0) - 2.0) < FLT_EPSILON);
+    TEST(fabs(wrap(-1.5, 3.0, 5.0) - 4.5) < FLT_EPSILON);
+    TEST(fabs(wrap(15.5, 3.0, 5.0) - 3.5) < FLT_EPSILON);
+    TEST(fabs(wrap(-1.0, 30.0, 40.0) - 39.0) < FLT_EPSILON);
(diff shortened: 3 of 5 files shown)


matrix/stdlib_imports.hpp
```diff
@@ -13,6 +13,14 @@
 #include <cstdlib>
 #include <cinttypes>
 
+#ifndef M_PI
+#define M_PI	3.14159265358979323846
```
dagar: Can we avoid defining our own PI? It's not standard, but I believe all of our platforms support _USE_MATH_DEFINES.

matrix/helper_functions.hpp
```diff
@@ -20,64 +20,39 @@ bool is_finite(Type x) {
...
+ * @param x input possibly outside of the range
+ * @param low lower limit of the allowed range
+ * @param high upper limit of the allowed range
+ * @return wrapped value indside the range
+ */
+template<typename Type>
+Type wrap(Type x, Type low, Type high) {
+    Type range = high - low;
+    x = fmod(x - low, range);
+    if (x < 0)
```
dagar: Would you mind using braces here? It helps prevent unintentional errors. Imagine someone adding another statement after `x += range;` with the same indentation.

matrix/helper_functions.hpp
```diff
@@ -20,64 +20,39 @@ bool is_finite(Type x) {
...
+ *
+ * @param x input possibly outside of the range
+ * @param low lower limit of the allowed range
+ * @param high upper limit of the allowed range
+ * @return wrapped value indside the range
+ */
+template<typename Type>
+Type wrap(Type x, Type low, Type high) {
+    Type range = high - low;
+    x = fmod(x - low, range);
```
dagar: Which fmod is this on NuttX?

In C fmod uses double precision input and out. fmodf uses floats.
In C++ fmod (std::fmod) has multiple overloads for different types (float, double, etc).

Accidentally promoting to double on an stm32f4 is a significant performance penalty that's even worse than the previous code.

dagar:
> Probably we can make it even more efficient by inlining which I suppose the compiler already does automatically. And adding a check if the value is already in range but I'm not sure if that really safes any compute.

The compiler already has everything it needs to inline this if it wants to. I'd strongly recommend profiling a system running something real before trying to optimize. The results are often not what you'd expect.

dagar:
Disassembly of matrix::wrap() (build for px4fmu-v4)
Note the fmod call.
image

Follow fmod and we see that it's ultimately calling __ieee754_fmod
image

dagar:
Quick little microbenchmark, again on a pixracer (no double floating point unit).

100,000 iterations of calling wrap_pi with random floats between 0 and 100.

image

Summary:
 updated wrap_pi: 

Now let's fix matrix stdlib_imports to ensure we're using fmodf and try again.

image


### Summary

| code        | per wrap_pi call (us)        | 100,000 wrap_pi calls (us)  |
| ------------- |:-------------:| -----:|
| wrap_pi Matrix master     | 2 us  |  915347 us |
| wrap_pi fmod (original #71 14d7394780eced6)  | 7 us | 1551347 us |
| wrap_pi fmodf (updated #71) 3071ec92e241ca9 | 3 us      |  1099219 us |


Surprising isn't it?

matrix/stdlib_imports.hpp
MaEtUgR: That was exactly my goal and when I run the tests on my computer it all works but on CI not. So I tried
```
#define _USE_MATH_DEFINES
#undef __STRICT_ANSI__
#include <cmath>
```
but CI still didn't find pi... So if you have any clue if we can change the CI call such that it works.

matrix/helper_functions.hpp
MaEtUgR: Sure sorry, I know the convention, I fell back to my old code style when reading through examples.
astyle enforces this on PX4...

matrix/helper_functions.hpp
MaEtUgR: You're totally right.

matrix/stdlib_imports.hpp
kcolton: Agree that rediscovering PI is not ideal and may be another good change, but isn't directly related to the PR's primary objective. To keep from blocking maybe a new branch?

matrix/helper_functions.hpp
```diff
@@ -20,64 +20,40 @@ bool is_finite(Type x) {
...
-
-    while (x < Type(-M_PI)) {
-        x += Type(2 * M_PI);
-
-        if (c++ > 100) {
-            return INFINITY;
-        }
+Type wrap(Type x, Type low, Type high) {
+    Type range = high - low;
+    x = fmod(x - low, range);
```
kcolton: Would `fmodf` be more appropriate here?

test/helper.cpp
```diff
@@ -5,18 +5,26 @@ using namespace matrix;
...
+    TEST(fabs(wrap_pi(-1001.0f) - (-1001.0f + 318 * M_PI)) < 1e-3);
+    TEST(!is_finite(wrap_pi(NAN)));
 
     TEST(fabs(wrap_2pi(-4.0) - (-4.0 + 2*M_PI)) < 1e-5);
     TEST(fabs(wrap_2pi(3.0) - (3.0)) < 1e-3);
-    TEST(!is_finite(wrap_2pi(1000.0f)));
-    TEST(!is_finite(wrap_2pi(-1000.0f)));
-    wrap_2pi(NAN);
+    TEST(fabs(wrap_2pi(2000.0f) - (2000.0f - 318 * M_TWOPI)) < 1e-3);
+    TEST(fabs(wrap_2pi(-2001.0f) - (-2001.0f + 319 * M_TWOPI)) < 1e-3);
```
kcolton: After performing the updated ops are all the data types still the same? Or would some vals be changing to doubles?

matrix/helper_functions.hpp
kcolton: oops - i missed the discussion in the issue. looks like it was already talked about and benchmarked. dont mind me :)

kcolton:
@dagar Awesome benchmarks. Curious how they scale if the range was larger or had negative numbers.

LorenzMeier:
If the range is larger then your code is super buggy as you will never have larger numbers for a PI wrap under normal conditions.

LorenzMeier:
The incorrect base assumption here is that this recreates a division. It’s not. Furthermore on a hardware level things that look like an atomic operator are in fact loops, often executed in parallel and with heuristics to arrive at the result in a fixed amount of time. And it depends very much on the architecture - so benchmarking is really the only way to argue optimality.

MaEtUgR:
Any opinions on this that would lead to merging or closing?

dagar:
Interested in measuring again?

MaEtUgR:
@dagar Sure. Do you still have your test bench branch? Then I can do that.

dagar:
You could put something in here - https://github.com/PX4/Firmware/blob/master/src/systemcmds/tests/test_microbench_matrix.cpp


	nsh> tests microbench_matrix
	INFO  [tests] RUNNING TEST: time_px4_matrix
	matrix Euler from Quaternion: 1000 events, 9106us elapsed, 9.11us avg, min 8us max 10us 0.314us rms
	matrix Euler from Dcm: 1000 events, 7659us elapsed, 7.66us avg, min 7us max 8us 0.474us rms
	matrix Quaternion from Euler: 1000 events, 9234us elapsed, 9.23us avg, min 9us max 10us 0.424us rms
	matrix Quaternion from Dcm: 1000 events, 2179us elapsed, 2.18us avg, min 2us max 3us 0.384us rms
	matrix Dcm from Euler: 1000 events, 9291us elapsed, 9.29us avg, min 9us max 10us 0.454us rms
	matrix Dcm from Quaternion: 1000 events, 2449us elapsed, 2.45us avg, min 2us max 3us 0.498us rms
	INFO  [tests] TEST PASSED: time_px4_matrix

	INFO  [tests] ALL TESTS PASSED
	INFO  [tests]   Tests passed :      1
	INFO  [tests]   Tests failed :      0
	INFO  [tests]   Tested assertions : 0
	microbench_matrix PASSED


Don't put too much weight on the actual timing numbers in these crude benchmarks, but they can help give you an idea of roughly what's going on.

LorenzMeier:
`fmod` should NOT be used for angles in an embedded system. We know that the angles are very few iterations off (1-2) if the system is in any sane condition. If you get 239393494 degrees as input you might as well just throw NaN, because you know that the input was invalid. Fast math code does not use boilerplate math functions. The same is true for `pow()` - you should NEVER EVER use that. Anyone using that for linear algebra doesn't understand anything about programming. And I'm saying that so pointed to make sure anyone finding this comment spends the time to read up on this and ask their co-workers.

MaEtUgR:
Sure, I'll finish this pr without it.

MaEtUgR:
I fixed it up and cleaned up the history. Anyone up for a review? We can use this for the wrapping in https://github.com/PX4/Firmware/pull/12883

matrix/helper_functions.hpp
```diff
@@ -20,64 +20,57 @@ bool is_finite(Type x) {
 #endif
 }
 
+/**
+ * Wrap value to stay in range [low, high)
+ *
+ * @param x input possibly outside of the range
+ * @param low lower limit of the allowed range
+ * @param high upper limit of the allowed range
+ * @return wrapped value inside the range
```
julianoes:
```suggestion
 * @return wrapped value inside the range, or NAN if value is too far away from range.
```

matrix/helper_functions.hpp
```diff
@@ -20,64 +20,57 @@ bool is_finite(Type x) {
...
+ */
 template<typename Type>
-Type wrap_pi(Type x)
-{
-    if (!is_finite(x)) {
+Type wrap(Type x, Type low, Type high) {
```
jkflying: Maybe we should have an integer overload version that uses the `%` operator? This is very fast on ARM...

Also, I would have phrased the API as `Type wrap(Type x, Type range, Type offset)`, so the wrap_2pi would be `wrap(x, M_TWOPI, 0)`, and wrap_pi would be `wrap(x, M_TWOPI, M_PI)`. But this one is fine too, just different ways of thinking about the problem.

PR #72, opened 2018-08-27
use default constructors and skip unnecessary initialization

dagar:
Saves ~ 7.7KB of flash in px4fmu-v3_default.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -24,7 +24,7 @@ namespace matrix
 template <typename Type, size_t M>
 class Vector;
 
-template<typename Type, size_t  M, size_t N>
+template<typename Type, size_t M, size_t N>
 class Matrix
 {
 
@@ -35,12 +35,12 @@ public:
     // Constructors
     Matrix() = default;
 
-    Matrix(const Type data_[][N])
+    Matrix(const Type data_[M*N])
     {
         memcpy(_data, data_, sizeof(_data));
     }
 
-    Matrix(const Type *data_)
+    Matrix(const Type data_[M][N])
     {
         memcpy(_data, data_, sizeof(_data));
     }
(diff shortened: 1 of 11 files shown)


dagar:
We might also want to consider making the various constructors explicit to avoid unexpected conversions.

matrix/Matrix.hpp
```diff
@@ -35,12 +35,12 @@ class Matrix
     // Constructors
     Matrix() = default;
 
-    Matrix(const Type data_[][N])
+    Matrix(const Type data_[M])
```
dagar: TODO review
 - Mx1 constructor (vector)
 - do we need to support MxN matrix construction from an array that's M*N elements

dagar:
Other than saving a surprising amount of flash I'm working towards closing all the memory safety holes.

matrix/Matrix.hpp
```diff
@@ -24,28 +24,28 @@ namespace matrix
...
 
-    Type _data[M][N];
+    Type _data[M][N] {};
 
     // Constructors
-    Matrix() : _data() {}
+    Matrix() = default;
 
-    Matrix(const Type data_[][N]) : _data()
+    Matrix(const Type data_[M])
```
bkueng: Do we need this one? If N!=1 you'll have an illegal memory access. Changing to `sizeof(data_)` would be fine too.

matrix/Scalar.hpp
```diff
@@ -17,21 +17,18 @@ template<typename Type>
 class Scalar
 {
 public:
-    Scalar() : _value()
-    {
-    }
+    Scalar() = default;
```
bkueng: Is this useful?

dagar: Nope, originally wasn't sure if we could get away with Scalar being constant, but it appears that we can.

matrix/Matrix.hpp
dagar: We do for now, a lot of the testing constructs a matrix from an array. Check the updated commit.

PR #75, opened 2018-10-01
Vector2 from Vector3 constructor

bkueng:
The PX4 Firmware (in Flight Tasks) contains code like:
```
Vector2f vel(&_velocity(0));
Vector2f vel_sp_xy = Vector2f(&_velocity_setpoint(0));
```

This is unnecessary hard to read. We can simplify the code by adding an explicit constructor for Vector2 from Vector3, so that we can write:
```
Vector2f vel(_velocity);
Vector2f vel_sp_xy(_velocity_setpoint);
```

@Stifael @MaEtUgR

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,15 @@ A simple and efficient template based matrix library.
 * matrix/integrate.hpp : Provides integration routines.
     * integrate_rk4 (Runge-Kutta 4th order)
 
+## Testing
+The tests can be executed as following:
+```
+mkdir build
+cd build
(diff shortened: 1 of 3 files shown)


LorenzMeier:
Awesome! Let's keep deleting code!

MaEtUgR:
There's also `Vector2f v1(&v2(0));` with `v2` being `Vector<Type, 4>`. Let's see how many there are to evaluate if it's worth another constructor.

PR #76, opened 2018-10-28
Add slicing tests that are not pure row slicing

madcowswe:
I had a look at the implementation of `slice`, and I found it odd that it doesn't have a copy loop. The current implementation does a raw memcpy of the underlying contiguous row-major data. As far as I can tell, this could only slice along rows. Interestingly, I found that the tests only tested for slicing along rows, so this bug would go unnoticed.

I added some tests that checks slicing along columns also. I have a feeling this would break, and we need to fix the implementation of `slice`. However I could be wrong, and hence I'm submitting these tests first to verify.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -360,7 +360,13 @@ public:
     template<size_t P, size_t Q>
     Matrix<Type, P, Q> slice(size_t x0, size_t y0) const
     {
-        Matrix<Type, P, Q> res(&(_data[x0][y0]));
+        Matrix<Type, M, N> &self = *this;
+        Matrix<Type, P, Q> res; //default constructed
+        for (size_t i = 0; i < P; i++) {
+            for (size_t j = 0; j < Q; j++) {
+                res(i, j) = self(i + x0, j + y0);
+            }
+        }
         return res;
     }
 
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -360,7 +360,7 @@ public:
     template<size_t P, size_t Q>
     Matrix<Type, P, Q> slice(size_t x0, size_t y0) const
     {
-        Matrix<Type, M, N> &self = *this;
+        const Matrix<Type, M, N> &self = *this;
         Matrix<Type, P, Q> res; //default constructed
         for (size_t i = 0; i < P; i++) {
             for (size_t j = 0; j < Q; j++) {
--- a/test/slice.cpp
+++ b/test/slice.cpp
@@ -31,7 +31,7 @@ int main()
     TEST(isEqual(B_colslice, B_check_colslice));
 
     // Test slicing both
-    Matrix<float, 3, 2> B_bothslice(A.slice<2, 2>(1, 1));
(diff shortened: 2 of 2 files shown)


madcowswe:
I had a look at the output of the tests on Travis, and indeed `slice` is broken.
I would suggest to replace the memcpy implementation with an explicit copy-loop.

jgoppert:
Thanks for finding this. Please format the code if you get a chance so travis passes.

madcowswe:
Sorry do you mean I should create the code for a working `slice` implementation, or do you mean that the formatting of my provided test-code is not adequate?

MaEtUgR:
@madcowswe nice catch! btw you can also run the tests on your machine, see https://github.com/PX4/Matrix#testing

TSC21:
I also did this test yesterday and found it was not working. I feel we need to get something similar to what the `Block` implementation is doing in Eigen.

jgoppert:
I feel like we could do a check if the fast row access is valid, if not, switch to a slower for loop approach.

madcowswe:
I don't see how using `memcpy` is faster than a copy loop on a platform like ARM Cortex, likely `memcpy` would actually be implemented as a copy loop. In fact, a copy loop on floats would be able to copy full 32-bit words at a time, while `memcpy` is done with bytes, and could therefore even be slower.

My recommendation is to discard the `memcpy` implementation and replace it with a 2D copy loop.

mhkabir:
@bkueng Thoughts on performance?

jgoppert:
Some info on memcpy: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka3934.html, we should probably just test on hardware. Either way, having a working for loop version should be priority to get this bug fixed.

madcowswe:
Ah yeah it looks like the compiler would be able to choose a 32 bit version of `memcpy`. Still, since this is a header library, it's likely that a copy loop would just be inlined or at least very optimised.

bkueng:
> Some info on memcpy: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka3934.html, we should probably just test on hardware. Either way, having a working for loop version should be priority to get this bug fixed.

I agree with that. `memcpy` is generally heavily optimized and uses the fastest instructions possible. However branching adds penalties too, and for the small matrices we are using it for, a loop is probably fine.

TSC21:
@jgoppert @bkueng @MaEtUgR can we agree then on an approach that helps fixing this?

madcowswe:
I implemented slice with a copy loop. It mirrors the implementation of `set` right below it. Seems to pass the tests. 
Though I don't know what the format error is about on the GCC build.

bkueng:
Can you check CI?

```
checking format...
Error: need to format
```

TSC21:
@madcowswe if you have time, please run `make format` and then commit again.

TSC21:
@MaEtUgR can you please fix rhe trailing ws then?

MaEtUgR:
Done: https://github.com/PX4/Matrix/commit/56b069956da141da244926ed7000e89b2ba6c731

TSC21:
Perfect

PR #80, opened 2019-02-17
Feature: add NaN set and URT of matrix

TSC21:
This PR adds support for:
- Setting a full matrix or vector with NANs;
- Get the upper-right triangle of a `SquareMatrix`.

I also did some improvements on the testing.

### DIFF
--- a/matrix/SquareMatrix.hpp
+++ b/matrix/SquareMatrix.hpp
@@ -149,7 +149,7 @@ bool inv(const SquareMatrix<Type, M> & A, SquareMatrix<Type, M> & inv)
     for (size_t n = 0; n < M; n++) {
 
         // if diagonal is zero, swap with row below
-        if (fabs(static_cast<float>(U(n, n))) < 1e-8f) {
+        if (fabs(static_cast<float>(U(n, n))) < __FLT_EPSILON__) {
             //printf("trying pivot for row %d\n",n);
             for (size_t i = n + 1; i < M; i++) {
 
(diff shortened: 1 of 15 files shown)


TSC21:
I already tried to run `make format` and states that all is good. So I don't understand why the format is failing here.

jgoppert:
Looks like you need a urt test case.

TSC21:
> Looks like you need a urt test case.

Yep I am working on that.

matrix/SquareMatrix.hpp
```diff
@@ -133,7 +149,7 @@ bool inv(const SquareMatrix<Type, M> & A, SquareMatrix<Type, M> & inv)
     for (size_t n = 0; n < M; n++) {
 
         // if diagonal is zero, swap with row below
-        if (fabs(static_cast<float>(U(n, n))) < 1e-8f) {
+        if (fabs(static_cast<float>(U(n, n))) < __FLT_EPSILON__) {
```
dagar:
```suggestion
        if (fabs(static_cast<float>(U(n, n))) < FLT_EPSILON) {
```

TSC21: Done. I add to define it so it could work.

TSC21:
@dagar all good now? It's already being used in https://github.com/PX4/Firmware/pull/11092.

TSC21:
I am merging this as tests pass, it has coverage, and it's already being used on PX4 codebase.

matrix/stdlib_imports.hpp
```diff
@@ -15,6 +15,10 @@
 
 namespace matrix {
 
+#if !defined(FLT_EPSILON)
+#define FLT_EPSILON     __FLT_EPSILON__
```
dagar: @TSC21 this is in `float.h`.

TSC21: Including `float.h` resulted on definition conflicts with the other platforms. Since we just need `FLT_EPSILON` definition, it made more sense to me just include this definiotn.

dagar: If including a standard c header caused conflicts there are other problems.

PR #81, opened 2019-08-22
Canonical Quaternion

kamilritz:
Added function that will get you the quaternion in canoncial form. A canonical quaternion has an non-negative real part. If it has a negative real part, multiplying every component with -1 will bring it to canonical form. Both quaternion will represent the same orientation.

### DIFF
--- a/matrix/Quaternion.hpp
+++ b/matrix/Quaternion.hpp
@@ -353,6 +353,30 @@ public:
     }
 
     /**
+     * Bring quaternion to canonical form
+     */
+    void canonicalize()
+    {
+        *this = this->canonical();
+    }
+
+
+    /**
+     * Return canonical form of the quaternion
+     *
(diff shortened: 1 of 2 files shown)


matrix/Quaternion.hpp
```diff
@@ -352,6 +352,30 @@ class Quaternion : public Vector<Type, 4>
...
+
+	/**
+	 * Return canonical form of the quaternion
+	 *
+	 * @return quaternion in canonical from
+	 */
+	Quaternion canonical() const
+	{
+		const Quaternion &q = *this;
+		if(q(0)<0.0f){
```
jkflying:
```suggestion
		if(q(0) < Type(0)){
```

kamilritz: Changed it

PR #82, opened 2019-08-26
matrix: add method to check all values are nan

mrivi:


matrix/Matrix.hpp
```diff
@@ -507,6 +507,17 @@ class Matrix
         return min_val;
     }
 
+    bool isAllNan() {
```
jkflying:
```suggestion
    bool isAllNan() const {
```

matrix/Matrix.hpp
```diff
@@ -507,6 +507,17 @@ class Matrix
...
+      Matrix<float, M, N> &self = *this;
```
jkflying:
```suggestion
      const Matrix<float, M, N> &self = *this;
```

test/matrixAssignment.cpp
```diff
@@ -30,6 +30,7 @@ int main()
     for(int i=0; i<9; i++) {
         TEST(isnan(m_nan.data()[i]));
     }
+    TEST(m_nan.isAllNan());
```
jkflying: Also add a test for the negative result, to make sure it doesn't always return `true`

PR #83, opened 2019-09-05
fix bug in matrix wrap function

baumanta:
There was a bug in the wrapping function which returned NAN for wrapping a value equal to the lower boundary. E.g wrap(0.0, 0.0, 360.0) = nan


MaEtUgR:
Thanks for fixing this! I got a follow up with more tests #84 because I got paranoid.

PR #85, opened 2019-09-05
Smaller code size wrap

jkflying:
After trying to bring the last wrap into the firmware, we discovered it was too big. It seems to have a disproportionate size due to inlining. This is a smaller version, that should still be fast.

Input arguments for low/high are in all cases I saw, const, so I explicitly inverted and multiplied the range in the floor so we don't have a divide due to compiler const propagation, see https://godbolt.org/z/RYZIoR
It also has a 'happy path' with no extra work, for the 99% case where things don't require wrapping.

This drops the codesize in FMUv2 by more than 700 bytes in my testing.

### DIFF
--- a/matrix/helper_functions.hpp
+++ b/matrix/helper_functions.hpp
@@ -26,7 +26,7 @@ bool is_finite(Type x) {
  * @param x input possibly outside of the range
  * @param low lower limit of the allowed range
  * @param high upper limit of the allowed range
- * @return wrapped value inside the range, or NAN if value is too far away from range.
+ * @return wrapped value inside the range
  */
 template<typename Type>
 Type wrap(Type x, Type low, Type high) {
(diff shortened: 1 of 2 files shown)


matrix/helper_functions.hpp
```diff
@@ -35,24 +35,10 @@ Type wrap(Type x, Type low, Type high) {
...
-        while (x < low) {
-            x += range;
-        }
-        return x;
-    }
-
-    // very far from the range -> something went terribly wrong
-    return NAN;
+    const Type range = high - low;
+    const Type inv_range = Type(1) / range; // should evaluate at compile time, multiplies below at runtime
```
MaEtUgR: > at compile time

with fixed range yes but's that's also the expected use

jkflying:
`fmod` does something super crazy: http://www.netlib.org/fdlibm/e_fmod.c

MaEtUgR:
I did some more practical tests this morning and the only difference in runtime I see is that the old method doesn't scale to a large number of wraps which was expected and why I wanted to change it. But with one wrap I don't see a difference in run time because it's below the small execution time jitter "noise".

PR #86, opened 2019-09-09
Replace pow with sqrt

jkflying:
Currently the only usage of `pow` on vector is with a constant argument of 0.5, for sqrt. This is inefficient and increases codesize.

This PR replaces the calls to pow with argument 0.5, with sqrt. Firmware PR [coming soon](https://github.com/PX4/Firmware/tree/pr-remove-calls-to-pow)

Proof GCC does not transform pow(x,0.5) into a sqrt instruction: https://godbolt.org/z/Cu7dQh

### DIFF
--- a/matrix/Vector.hpp
+++ b/matrix/Vector.hpp
@@ -66,7 +66,7 @@ public:
 
     Type norm() const {
         const Vector &a(*this);
-        return Type(sqrt(a.dot(a)));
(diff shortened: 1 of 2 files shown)


MaEtUgR:
Good find! how much flash did you safe? 😉

jkflying:
Maybe it's worse since more things inline now...

PR #89, opened 2019-09-12
Write to slices

jkflying:
This PR is inspired by the Eigen API, which has a writable `.block()` function.

In order to achieve this, I've made a Slice which acts as a temporary, to decide how data assigned to/from the `.slice()` function is treated.

Testing in the PX4 firmware, it compiles binaries to exactly the same size as before despite the use of the additional temporary struct.

### DIFF
--- a/test/vector3.cpp
+++ b/test/vector3.cpp
@@ -27,6 +27,29 @@ int main()
     TEST(isEqual(a.unit(), a.normalized()));
     TEST(isEqual(a*2.0, Vector3f(2, 0, 0)));
 
+    Vector2f g2(1,3);
+    Vector3f g3(7, 11, 17);
+    g3.xy() = g2;
+    TEST(isEqual(g3, Vector3f(1, 3, 17)));
+
+    const Vector3f g4(g3);
+    Vector2f g5 = g4.xy();
+    TEST(isEqual(g5,g2));
+    TEST(isEqual(g2,Vector2f(g4.xy())));
+
+    Vector3f h;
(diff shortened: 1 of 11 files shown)


test/vector3.cpp
```diff
@@ -27,6 +27,19 @@ int main()
     TEST(isEqual(a.unit(), a.normalized()));
     TEST(isEqual(a*2.0, Vector3f(2, 0, 0)));
 
+    Vector2f g2(1,3);
+    Vector3f g3(7, 11, 17);
+    g3.xy() = g2;
```
jkflying: @bresch how does this look?

jkflying:
@dagar Coveralls seems to be broken, any idea where to start looking?

jgoppert:
https://coveralls.io/github/PX4/Matrix, seems to have completed there

jkflying:
So we're just not getting the update here then. Weird.

test/vector3.cpp
bresch: That's awesome! Thanks a lot. After that gets available in the firmware, I'll make a pass to get rid of the 1-by-1 assignments for xy axes.

jkflying:
Ok, I'm pretty happy with this now. Let's get this in, I'd like to get rid of the raw data accesses (the `.data()` functions) next and then work on lazy evaluation to support things like `.xy().norm()` and `float res = A.row(i).dot(B.col(j));` without performance penalties above  the raw data access.

PR #90, opened 2019-09-16
Remove direct access to internal data

jkflying:
This PR removes the `.data()` function which allowed unsafe access to the internal data structures of Matrix.

@dagar


dagar:
Needs to be rebased before github will "Rebase and merge".

PR #92, opened 2019-09-16
Map: wrap pointers safely

jkflying:
This PR brings in a rough approximation of the Eigen concept of [Map](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html), allowing to wrap pointers safely and assign structures (Matrix, Vector, Slice) to or from the memory underlying. The idea is that this can replace the `copyTo()` function in a more idiomatic manner, with less temporaries required due to direct interaction with Slices.

### DIFF
--- a/matrix/Map.hpp
+++ b/matrix/Map.hpp
@@ -33,7 +33,7 @@ public:
 
     Map(Type data[M][N])
     {
-        _data = data;
+        _data = reinterpret_cast<Type*>(data);
     }
 
     Map<Type, M, N> & operator=(const Matrix<Type, M, N> &other)
--- a/matrix/Map.hpp
+++ b/matrix/Map.hpp
@@ -1,0 +1,72 @@
+/**
+ * @file Map.hpp
+ *
+ * Map class, for wrapping chunks of memory without taking ownership or copying
+ *
+ * @author Julian Kent <julian@auterion.com>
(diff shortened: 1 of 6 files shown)


jkflying:
This is perhaps more complicated than just adding a `copyTo` function to Slice and any other types we have in the future. Unless people think this is a better way to go I'll close this.

MaEtUgR:
I like the idea of safely and expressively being able to assign to parts of a matrix since that is used a lot in the controllers and also setpoint generation. But I'm not fully aware of all the things necessary to realize it. It might be more efficient to have specific implementations for Vector2 and Vector3 because that's where it's used most. What is your take away after these tries?

jkflying:
It depends where we want to go with this. If we want to be able to take external memory and directly use it to multiply a matrix, or add to a matrix etc., *without* copying it to a Matrix class first, then we need something like this Map class. If all we will ever use it for is copying data to/from a Matrix, then I think we could put our time better elsewhere.

MaEtUgR:
The only use case I usually run into is the `msg`/uORB topic containing an array or even separate float values. So I copy into a Matrix/Vector, process and copy to the next topic. But I'm not sure it's worth to improve that further because there are more burning things.

jkflying:
Let's leave this for later then.

PR #93, opened 2019-09-17
Fix equality checks, NAN being equal to any number

MaEtUgR:
While writing unit tests for position control in https://github.com/PX4/Firmware/pull/12072 I had tests passing when I intentionally broke them. I tracked it down to the matrix library reporting `a == b` being `true` even though the elements were different if on either side there was a `NAN` in that specific position.

This pr fixes the problem, adds tests for it and resolves all the issues that were covered by tests that passed even though they should have failed.

### DIFF
--- a/matrix/helper_functions.hpp
+++ b/matrix/helper_functions.hpp
@@ -34,7 +34,9 @@ bool is_finite(Type x) {
 template<typename Type>
 bool isEqualF(const Type x, const Type y, const Type eps = 1e-4f)
 {
-    return matrix::fabs(x - y) <= eps;
+    return (matrix::fabs(x - y) <= eps)
+           || (isnan(x) && isnan(y))
+           || (isinf(x) && isinf(y));
 }
 
 /**
--- a/matrix/helper_functions.hpp
+++ b/matrix/helper_functions.hpp
@@ -23,8 +23,8 @@ bool is_finite(Type x) {
 /**
  * Compare if two floating point numbers are equal
  *
- * Note: Smaller or EQUAL than is important to correctly
- * handle the comparison to infinite or nan.
+ * NAN is considered equal to NAN and -NAN
+ * INFINITY is considered equal INFINITY but not -INFINITY
  *
  * @param x right side of equality check
  * @param y left side of equality check
@@ -36,7 +36,7 @@ bool isEqualF(const Type x, const Type y, const Type eps = 1e-4f)
 {
     return (matrix::fabs(x - y) <= eps)
            || (isnan(x) && isnan(y))
-           || (isinf(x) && isinf(y));
+           || (isinf(x) && isinf(y) && isnan(x - y));
 }
 
 /**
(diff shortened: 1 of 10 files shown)


MaEtUgR:
We actually never had proper unit testing for the print functions, they were just called in a random place before.

test/helper.cpp
```diff
@@ -39,13 +39,27 @@ int main()
...
     TEST(!isEqual(a, b));
     TEST(isEqual(a, a));
 
-    TEST(isEqualF(1., 1.));
-    TEST(!isEqualF(1., 2.));
+    Vector3f c(1, 2, 3);
+    Vector3f d(1, 2, NAN);
+    TEST(!isEqual(c, d));
+    TEST(isEqual(c, c));
+    TEST(!isEqual(d, d));
```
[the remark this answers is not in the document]
MaEtUgR: Good point. I considered it and then thought we are "forced by convention" to follow floating-point arithmetic. But it would simplify client code a lot if you could use the equal check for NAN as well if you expect it as output.

MaEtUgR:
I'm writing a unit test for the output of matrices since we never had one.

test/helper.cpp
```diff
@@ -46,8 +46,9 @@ int main()
     TEST(!isEqualF(1.f, NAN));
     TEST(!isEqualF(INFINITY, 1.f));
     TEST(!isEqualF(1.f, INFINITY));
-    TEST(!isEqualF(NAN, NAN));
-    TEST(!isEqualF(INFINITY, INFINITY));
+    TEST(isEqualF(NAN, NAN));
+    TEST(isEqualF(NAN, -NAN));
+    TEST(isEqualF(INFINITY, INFINITY));
```
jkflying: What about 
`TEST(!isEqualF(INFINITY, -INFINITY));`

MaEtUgR: Sure, wait a mom.

MaEtUgR: How should I check that if I can't use == with the float? -INFINITY !< 0

MaEtUgR: I found an efficient way 🎉

MaEtUgR: https://github.com/PX4/Matrix/pull/93/commits/e9157f71a350c0df3ed31c6d00935209508e947c

matrix/helper_functions.hpp
```diff
@@ -20,6 +20,23 @@ bool is_finite(Type x) {
...
+ *
+ * @param x right side of equality check
+ * @param y left side of equality check
+ * @param eps numerical tolerance of the check
+ * @return true if the two values are considered equal, false otherwise
+ */
+template<typename Type>
+bool isEqualF(const Type x, const Type y, const Type eps = 1e-4f)
+{
+    return matrix::fabs(x - y) <= eps;
```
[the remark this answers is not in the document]
MaEtUgR: Do I need to change anything?

test/helper.cpp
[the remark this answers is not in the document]
MaEtUgR: https://github.com/PX4/Matrix/pull/93/commits/2fe6ec655d019382c327529f06bc7d5abfb0f787

PR #94, opened 2019-09-17
Don't lose array sizes in copyTo

jkflying:
Before, if the array of fixed size was passed as a function argument, it would instead be treated as a pointer since we are taking a reference of it. This fixes that, allowing an array to be passed into a function and then a Matrix.copyTo() to work.


MaEtUgR:
Cool! I didn't know that trick and was assuming if you don't pass the reference it is doing another copy in between...

PR #95, opened 2019-09-17
Fix type for division.

jgoppert:
This fixes a missing type promotion.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -270,7 +270,7 @@ public:
     void operator/=(Type scalar)
     {
         Matrix<Type, M, N> &self = *this;
-        self = self * (1.0f / scalar);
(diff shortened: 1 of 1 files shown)


matrix/Matrix.hpp
```diff
@@ -270,7 +270,7 @@ class Matrix
     void operator/=(Type scalar)
     {
         Matrix<Type, M, N> &self = *this;
-        self = self * (1.0f / scalar);
+        self = self * (Type(1.0f) / scalar);
```
MaEtUgR: Shouldn't it be `Type(1)` then? In case it's used as double template it doesn't really make sense to start out with a float.

MaEtUgR: I made a follow up for this: https://github.com/PX4/Matrix/pull/96

PR #97, opened 2019-09-30
More features: longerThan, norm, copyTo and slice on a Slice

jkflying:
This adds a few utility functions to make some common operations easier.

* ` vector.longerThan(float)` and `slice.longerThan(float)`. This is useful for checks, to hide the more efficient comparison of squared values while not making the code ugly. eg. `if (vel.xy().longerThan(VEL_HOR_LIMIT)) { ... } `
* `norm()` and `norm_squared()` added to Slice. For things like `vector3.xy().norm()` or `matrix3f.row(0).norm()`
* `copyTo(float[])` for a Slice. Don't need a matrix temporary to do `matrix3f.row(3).copyTo(buf)`
* Take a slice of a slice. Some things are easier to think about as sub-blocks, rather than absolute coordinates. Eg `Vector3f z_accel_variances = matrix<float,25,25>.slice<3,3>(6,6).slice<3,1>(1,0)`. This is mostly to make it easier to understand.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -90,12 +90,12 @@ public:
         return (*this);
     }
 
-    void copyTo(Type dst [M*N]) const
+    void copyTo(Type dst[M*N]) const
     {
         memcpy(dst, _data, sizeof(Type)*M*N);
     }
 
-    void copyToColumnMajor(Type dst [M*N]) const
+    void copyToColumnMajor(Type dst[M*N]) const
     {
         const Matrix<Type, M, N> &self = *this;
 
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -95,7 +95,7 @@ public:
         memcpy(dst, _data, sizeof(Type)*M*N);
     }
 
-    void copyToColumnMajor(Type (&dst)[M*N]) const
+    void copyToColumnMajor(Type dst [M*N]) const
     {
         const Matrix<Type, M, N> &self = *this;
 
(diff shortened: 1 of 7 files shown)


matrix/Slice.hpp
```diff
@@ -41,38 +41,95 @@ class Slice {
...
-                self(i, j) = in_matrix(i, j);
+                self(i, j) = other(i, j);
             }
         }
         return self;
     }
 
     // allow assigning vectors to a slice that are in the axis
-    Slice<Type, 1, Q, M, N>& operator=(const Vector<Type, Q>& in_vector)
+    template <size_t DUMMY = 1> // make this a template function since it only exists for some instantiations
```
MaEtUgR: I don't exactly see why this `DUMMY` template is necessary.

matrix/Matrix.hpp
```diff
@@ -95,7 +95,7 @@ class Matrix
         memcpy(dst, _data, sizeof(Type)*M*N);
     }
 
-    void copyToColumnMajor(Type (&dst)[M*N]) const
+    void copyToColumnMajor(Type dst [M*N]) const
```
MaEtUgR:
```suggestion
    void copyToColumnMajor(Type dst[M*N]) const
```
or is this a style preference?

matrix/Slice.hpp
jkflying: If you try to declare a fixed version like `template class matrix::Slice<float, 2, 3, 4, 5>;` that doesn't have 1 row, without this you get an error.

jkflying: Normally you'd fix this with `std::enable_if` but we don't have `<type_traits>` available.

PR #98, opened 2019-09-30
Enable branch coverage

jkflying:
This enables branch coverage in the coverage report


jkflying:
> Is the branch coverage somehow visualized in the web UI?

No, but you can see it locally. I got the branch coverage working with coveralls-lcov, so maybe we should try that instead.

PR #99, opened 2019-10-03
Make all array constructors explicit

MaEtUgR:
to avoid accidental implicit casts like e.g.
`Vector3f v = 0;`
assigning nullpointer content.

I have the Firmware pr ready to account for this restriction.

I still don't understand why the compiler implicitly allows above the above assignment which leads to an immediate segfault. It even allows it if the Vector constructors are not explicit but only the Matrix one they call. So you can wrap and explicit constructor in a non-explicit inheriting one...

Something remaining I still don't understand is the array size not getting checked:
```
float a[6] = {};
Vector3f v(a);
```
compiles just fine. I assume the intent of @dagar were compile time checks for array sizes in https://github.com/PX4/Matrix/pull/72

### DIFF
--- a/matrix/AxisAngle.hpp
+++ b/matrix/AxisAngle.hpp
@@ -37,7 +37,7 @@ public:
      *
      * @param data_ array
      */
-    AxisAngle(const Type data_[3]) :
+    explicit AxisAngle(const Type data_[3]) :
         Vector<Type, 3>(data_)
     {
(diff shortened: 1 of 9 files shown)


MaEtUgR:
Here's the Firmware pr fixing existing implicit casts: https://github.com/PX4/Firmware/pull/13088

PR #100, opened 2019-10-15
Automatic Differentiation 'Dual' Type

jkflying:
This adds the ability to do autodifferentiation. This is a high-perfomance method of calculating derivatives, partial derivatives and jacobians without depending on code generation or numerical differences. It gives accuracy and speed similar to generated code, but allows things not usually allowed in code generators like `if` blocks, `for` loops, discontinuities such as `min`/`max`/`fmod`, as well as the usual `sin`/`cos`/`asin`/`tan`/`atan`/`sqrt` etc.

Simple usage is in the tests, and here:

Eg given a function  **f(x,y,z) = x^2 + 2xy + 3y^2 + z** which we want the partial derivatives _d/dx_ and _d/dy_ at the point **(0.5, -0.8, 2)**

We just define the function as if we are evaluating the regular result, templated on the type in the vector.
```C++
template <typename T>
T testFunction(const Vector<T, 3>& point) {
    // function is f(x,y,z) = x^2 + 2xy + 3y^2 + z
    return point(0)*point(0)
           + 2.f * point(0) * point(1)
           + 3.f * point(1) * point(1)
           + point(2);
}
```
Normally to evaluate this functions we'd do something like:

```C++
Vector3<float> point(0.5f, -0.8f, 2.f);
float result = testFunction(point);
```

To get the partial derivatives, we simply evaluate it with a Vector with type `Dual<float, 2>` instead of `float`. The only 'tricky' part is that we have to specify the partial derivatives of our inputs so that it starts correctly, but this isn't that hard since _d/dx_ of **x** is 1, and _d/dy_ of **y** is also 1, by definition, since we want the derivatives with respect to the input.

```C++
using D = Dual<float, 2>;
// set our starting point, requesting partial derivatives of x and y in column 0 and 1
Vector3<D> dualPoint(D(0.5f, 0), D(-0.8f, 1), D(2.f));

Dual<float, 2> dualResult = testFunction(dualPoint);
float result = dualResult.value;
float d_dx = dualResult.derivative(0); // retrieve partial derivative of x in column 0
float d_dy = dualResult.derivative(1); // retrieve partial derivative of y in column 1
```
As can be seen this is extremely simple and easy to get partial derivatives.

This can be extended to work with any size of result, including vectors, matrices, quaternions etc, with any intermediate code such as matrix inversions, normalizations, and more, inside the function having any necessary logic for enabling terms, loops over dimensions, checks for normalization, etc. which code generators typically cannot handle.

Recommended reading:

-  [Automatic Differentiation, C++ Templates and Photogrammetry, by Dan Piponi](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.89.7749&rep=rep1&type=pdf)
-  http://ceres-solver.org/automatic_derivatives.html#implementing-jets

### DIFF
--- a/matrix/Dual.hpp
+++ b/matrix/Dual.hpp
@@ -1,0 +1,266 @@
+/**
+ * @file Dual.hpp
+ *
+ * This is a dual number type for calculating automatic derivatives.
+ *
+ * Based roughly on the methods described in:
+ * Automatic Differentiation, C++ Templates and Photogrammetry, by Dan Piponi
+ * and
+ * Ceres Solver's Jet.h
+ *
+ * @author Julian Kent <julian@auterion.com>
+ */
+
+#pragma once
+
+#include "math.hpp"
+
+namespace matrix
+{
+
+template <typename Scalar, size_t N>
+struct Dual
+{
+    static constexpr size_t WIDTH = N;
+
+    Dual() = default;
+
+    explicit Dual(Scalar v, size_t inputDimension = 65535)
+    {
+        value = v;
+        if (inputDimension < N) {
+            derivative[inputDimension] = Scalar(1);
+        }
+    }
+
+    explicit Dual(Scalar v, const Vector<Scalar, N>& d) :
+        value(v), derivative(d)
+    {}
+
+    Scalar value {};
+    Vector<Scalar, N> derivative;
+
+};
+
+// operators
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator+(const Dual<Scalar, N>& a)
+{
+    return a;
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator-(const Dual<Scalar, N>& a)
+{
+    return Dual<Scalar, N>(-a.value, -a.derivative);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator+(const Dual<Scalar, N>& a, const Dual<Scalar, N>& b)
+{
+    return Dual<Scalar, N>(a.value + b.value, a.derivative + b.derivative);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator-(const Dual<Scalar, N>& a, const Dual<Scalar, N>& b)
+{
+    return a + (-b);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator+(const Dual<Scalar, N>& a, Scalar b)
+{
+    return Dual<Scalar, N>(a.value + b, a.derivative);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator-(const Dual<Scalar, N>& a, Scalar b)
+{
+    return a + (-b);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator+(Scalar a, const Dual<Scalar, N>& b)
+{
+    return Dual<Scalar, N>(a + b.value, b.derivative);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator-(Scalar& a, const Dual<Scalar, N>& b)
+{
+    return a + (-b);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator*(const Dual<Scalar, N>& a, const Dual<Scalar, N>& b)
+{
+    return Dual<Scalar, N>(a.value * b.value, a.value * b.derivative + b.value * a.derivative);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator*(const Dual<Scalar, N>& a, Scalar b)
+{
+    return Dual<Scalar, N>(a.value * b, a.derivative * b);
+}
+
+template <typename Scalar, size_t N>
+Dual<Scalar, N> operator*(Scalar a, const Dual<Scalar, N>& b)
+{
+    return b * a;
+}
+
+template <typename Scalar, size_t N>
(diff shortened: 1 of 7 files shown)


jkflying:
CI failure is a segfault in the compiler, GCC 4.8

dagar:
> CI failure is a segfault in the compiler, GCC 4.8

Within PX4/Firmware only the armhf toolchain (raspberry pi) is this far back. Seems like a good excuse to finally update it. We occasionally hit other little issues with it.

jkflying:
Just as another example usecase, which I've added a test for. Say we have the following function:
```C++
template <typename Scalar>
Vector<Scalar, 3> positionError(const Vector<Scalar, 3>& positionState,
                                const Vector<Scalar, 3>& velocityStateBody,
                                const Quaternion<Scalar>& bodyOrientation,
                                const Vector<Scalar, 3>& positionMeasurement,
                                Scalar dt
                               )
{
    return positionMeasurement - (positionState +  bodyOrientation.conjugate(velocityStateBody) * dt);
}
```

And we want to get the jacobian of the position error (returned) with respect to the velocityStateBody (an input).

A (slow, inaccurate) numerical derivative would be like this:
```C++
Vector3f positionState(5,6,7);
Vector3f velocityState(-1,0,1);
Quaternionf velocityOrientation(0.2f,-0.1f,0,1);
Vector3f positionMeasurement(4.5f, 6.2f, 7.9f);
float dt = 0.1f;

Vector3f direct_error = positionError(positionState,
                                velocityState,
                                velocityOrientation,
                                positionMeasurement,
                                dt);

SquareMatrix3f numerical_jacobian;
for (size_t i = 0; i < 3; i++) // get finite differences for each dimension of the input
{
    Vector3f h3;
    float h = 0.001f;
    h3(i) = h;
    numerical_jacobian.col(i) = (positionError(positionState,
                                    velocityState + h3,
                                    velocityOrientation,
                                    positionMeasurement,
                                    dt)
                                    - direct_error)/h;
}
```

With auto-derivatives, it is like this:
```C++
using D = Dual<float, 3>;
using Vector3f3d = Vector3<D>;
Vector3f3d positionState(D(5), D(6), D(7));
Vector3f3d velocityState(D(-1), D(0), D(1));
Quaternion<D> velocityOrientation(D(0.2f),D(-0.1f),D(0),D(1));
Vector3f3d positionMeasurement(D(4.5f), D(6.2f), D(7.9f));
D dt(0.1f);

// get partial derivatives of velocity state:
velocityState(0).derivative(0) = 1;
velocityState(1).derivative(1) = 1;
velocityState(2).derivative(2) = 1;

Vector3f3d error = positionError(positionState,
                                    velocityState,
                                    velocityOrientation,
                                    positionMeasurement,
                                    dt);

Vector3f auto_error = Vector3f(error(0).value, error(1).value, error(2).value);
SquareMatrix3f auto_jacobian;
auto_jacobian.row(0) = error(0).derivative;
auto_jacobian.row(1) = error(1).derivative;
auto_jacobian.row(2) = error(2).derivative;
```

@RomanBapst this would be an alternative to the code generation in ECL.

jgoppert:
I've been using Casadi in my work for A/D. This is an interesting approach. What do you see as the advantages and disadvantages vs. using Casadi for code generation? @jkflying

jgoppert:
I also agree about code generation. It is better if we can do it at build time and not require a license. Here is my current approach, casadi based attitude estimator. Thanks for your input above. https://github.com/jgoppert/Firmware/tree/cei/src/modules/cei

matrix/Dual.hpp
```diff
@@ -0,0 +1,361 @@
...
+template <typename Scalar, size_t N>
+Dual<Scalar, N> atan2(const Dual<Scalar, N>& a, const Dual<Scalar, N>& b)
+{
+    // derivative is equal to that of atan(a/b), so substitute a/b into atan and simplify
+    Scalar atan_d = Scalar(1) / (a.value * a.value + b.value * b.value);
+    return Dual<Scalar, N>(atan2(a.value, b.value), (a.derivative * b.value - a.value * b.derivative) * atan_d);
+}
+
+template <typename Scalar, size_t M, size_t N>
+Matrix<Scalar, M, N> derivative(const Matrix<Dual<Scalar, N>, M, 1>& input)
```
kamilritz: This function has the same name as one of the member variables. Maybe we can chance this function name to something like "getDerivative"

matrix/Dual.hpp
```diff
@@ -0,0 +1,361 @@
...
+struct Dual
+{
+    static constexpr size_t WIDTH = N;
+
+    Scalar value {};
+    Vector<Scalar, N> derivative;
+
+    Dual() = default;
+
+    explicit Dual(Scalar v, size_t inputDimension = 65535)
```
kamilritz: It is not fully clear to me what is happening in this constructor. Can you give an example?

jkflying: Good catch, I've actually remove the one helper function, since it is better expressed using this constructor.

jkflying:
> Here is my current approach, casadi based attitude estimator. Thanks for your input above. https://github.com/jgoppert/Firmware/tree/cei/src/modules/cei

That looks interesting!

In comparison, this method doesn't need a consistent interface to get your auto-derivatives, you just use a different Scalar type and can mix it in wherever you need. Also, the code is human-readable, so can be verified for correctness. On the other hand, if your compiler isn't optimizing away the x+0 or x*1 then this will be a lot slower, so I do see the use for generated code in certain cases.

jkflying:
Found a workaround for the GCC segfault, which was mentioned [here](https://stackoverflow.com/a/51435283/1436711)

Checking that the workaround isn't a performance regression (although the fallback for non-IEEE floats is): https://godbolt.org/z/QZ02uJ

PR #101, opened 2019-11-01
Fix weird C preprocessor conflicts

jkflying:
Solution for https://github.com/PX4/Firmware/pull/13336

Unfortunately this is non-standard naming, so it isn't quite as find-replace as the old way. Hopefully gcc upgrades will let us go back to the old way.


jkflying:
(merging to take load off @dagar)

PR #103, opened 2019-11-04
Work around GCC-4.8 bug

jkflying:
In addition to updating the toolchain, this fixes the GCC 4.8 bug directly.

For reference:
https://github.com/PX4/Firmware/pull/13336
https://github.com/PX4/rpi_toolchain/pull/1

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -18,6 +18,14 @@
 
 #include "math.hpp"
 
+// There is a bug in GCC 4.8, which causes the compiler to segfault due to array {} constructors.
+// Do for-loop constructors just for GCC 4.8
+#ifdef __GNUC__
+#define MATRIX_GCC_4_8_WORKAROUND (__GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 9))
+#else
+#define MATRIX_GCC_4_8_WORKAROUND 0
+#endif
+
 namespace matrix
 {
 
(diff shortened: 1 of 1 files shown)


matrix/Matrix.hpp
```diff
@@ -33,24 +41,27 @@ class Slice;
 template<typename Type, size_t M, size_t N>
 class Matrix
 {
-
+#if MATRIX_GCC_4_8_WORKAROUND
```
MaEtUgR: I'm against having compiler specific "ifdefs" in the middle of the code. It increases complexity and could lead to untested corner cases. If it's only a redefinition I would totally accept it but executing different code in a math library seems not like a desired solution IMHO...

jkflying: I agree, I don't like it either. This is just a temporary workaround until the toolchain for Raspberry Pi can be updated to GCC 4.9, which doesn't segfault when trying to compile the `_data {}` initialization. Then we can update the compiler here to something more recent and get around this nasty hack. If you have a better idea I'm happy to do that instead.

Maybe this would be the right time to add some different versions of GCC to the CI as well, but already in the CI we are testing the GCC 4.8 as well as non-GCC codepaths, so at least both sides are being compiled/executed.

The reason it didn't work before was a big increase in binary size when not using the `=default` constructor.

dagar:
Let's move forward with this to get out of the way for other work (https://github.com/PX4/Matrix/pull/102), but make sure we come back to drop it as soon as possible.

PR #105, opened 2019-11-08
Add min, max, constrain function for Matrix

kamilritz:
This PR adds the functionality to compute a constrained Matrix with a scalar or another Matrix of the same size. It provides the functionality of min, max and constrain.

@jkflying

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -580,16 +580,50 @@ bool isEqual(const Matrix<Type, M, N> &x,
     return true;
 }
 
+namespace typeFunction
+{
+template<typename Type>
+Type min(const Type x, const Type y) {
+    bool x_is_nan = isnan(x);
+    bool y_is_nan = isnan(y);
+    // z > nan for z != nan is required by C the standard
+    if(x_is_nan || y_is_nan) {
+        if(x_is_nan && !y_is_nan) return y;
+        if(!x_is_nan && y_is_nan) return x;
+        return x;
+    }
+    return (x < y) ? x : y;
+}
+template<typename Type>
+Type max(const Type x, const Type y) {
+    bool x_is_nan = isnan(x);
+    bool y_is_nan = isnan(y);
+    // z > nan for z != nan is required by C the standard
+    if(x_is_nan || y_is_nan) {
+        if(x_is_nan && !y_is_nan) return y;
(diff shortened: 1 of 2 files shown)


matrix/Matrix.hpp
```diff
@@ -580,6 +580,126 @@ bool isEqual(const Matrix<Type, M, N> &x,
...
+                    m(i,j) = x_lower_bound(i,j);
+                } else if(x(i,j) > x_upper_bound(i,j)) {
+                    m(i,j) = x_upper_bound(i,j);
+                } else {
+                    m(i,j) = x(i,j);
+                }
+            }
+        }
+    }
+    if(should_be_nan) {
```
jkflying: Shouldn't we only set NaN to the values where this is true?

jkflying:
Something I just realised, the semantics of min and max need to match the standard library definition when one of the values is a NaN. This should be tested, so that eg. `min(NAN, 1) == min(Vector2f(NaN,5),1.f)(0)`, and vice versa with argument ordering.

kamilritz:
> Something I just realised, the semantics of min and max need to match the standard library definition when one of the values is a NaN. This should be tested, so that eg. min(NAN, 1) == min(Vector2f(NaN,5),1.f)(0), and vice versa with argument ordering.

@jkflying which implementation would you suggest for the reference min(NAN,1) call?

jkflying:
Preferably `fmin` and `fmax`, but `std::min`/`max` should have the same semantics.

matrix/Matrix.hpp
```diff
@@ -580,6 +580,128 @@ bool isEqual(const Matrix<Type, M, N> &x,
...
+    return (x < y) ? x : y;
+}
+template<typename Type>
+Type max(const Type x, const Type y) {
+    bool x_is_nan = isnan(x);
+    bool y_is_nan = isnan(y);
+    // z > nan for z != nan is required by C the standard
+    if(x_is_nan || y_is_nan) {
+        if(x_is_nan && !y_is_nan) return y;
+        if(!x_is_nan && y_is_nan) return x;
```
MaEtUgR: The second condition does the same as default hence it could be left away e.g. with a comment.

kamilritz: Good observation

MaEtUgR: https://github.com/PX4/Matrix/pull/126

PR #109, opened 2019-11-26
Vector3f cross product directly return result

dagar:


jkflying:
Does this actually change the resulting binary? I would have thought that the compiler would have optimized it.

dagar:
Yes, it does a slightly better job inlining everything, at least at `-Os`.

For example in MatrixTest. https://github.com/PX4/Firmware/blob/30fcb141563d5477450a55d575fcb832cf11bd31/src/systemcmds/tests/test_matrix.cpp#L676-L683

### Master

Screenshot from 2019-11-26 13-18-54



### PR

Screenshot from 2019-11-26 13-18-11

PR #110, opened 2019-11-28
Quaternion multiply inline return

dagar:


dagar:
Yes, although the areas that changed aren't really that hot. I suggest we do some profiling and look for opportunities in the inner loop and estimator.

PR #115, opened 2019-12-06
Asserts

kamilritz:
Adding assorts that check in debug mode if function arguments respect the size of the matrix


kamilritz:
I can confirm that it builds with ARM toolchain. tested with fmu v2

PR #116, opened 2019-12-18
Catch quaternion canonical corner cases

kamilritz:
@MaEtUgR Thanks for showing me these corner cases.


MaEtUgR:
Thanks @kamilritz , I made a pr to take advantage in PX4: https://github.com/PX4/Firmware/pull/14374

PR #117, opened 2019-12-26
Fix AngleAxis constructors

kamilritz:
The existing implementation of AxisAngle(Quaternion) caused the strange behavior in SITL flights in https://github.com/PX4/ecl/pull/696. The new implementation is known to be numerically more robust and is solving this issue.

Two other constructors were missing the call to AxisAngle(Quaternion)

Good reference for formulas: https://github.com/ANYbotics/kindr/blob/master/doc/cheatsheet/cheatsheet_latest.pdf

### DIFF
--- a/matrix/AxisAngle.hpp
+++ b/matrix/AxisAngle.hpp
@@ -69,16 +69,11 @@ public:
     AxisAngle(const Quaternion<Type> &q)
     {
         AxisAngle &v = *this;
-        Type ang = Type(2) * acos(q(0));
-        Type mag = sin(ang/2.0f);
-        if (fabs(mag) > 0) {
-            v(0) = ang*q(1)/mag;
-            v(1) = ang*q(2)/mag;
-            v(2) = ang*q(3)/mag;
+        Type mag = Type(sqrt(q(1) * q(1) + q(2) * q(2) + q(3) * q(3)));
+        if (fabs(mag) >= Type(1e-10)) {
+            v = q.imag() * Type(Type(2) * atan2(mag, q(0)) / mag);
         } else {
-            v(0) = 0;
-            v(1) = 0;
-            v(2) = 0;
+            v = q.imag() * Type(Type(2) * float(sign(q(0))));
         }
     }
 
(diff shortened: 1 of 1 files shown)


LorenzMeier:
I see weird twitching in SITL on boot up - is this related in any way to it?

kamilritz:
Probably not. The mentioned constructors were previously only used [once](https://github.com/PX4/Firmware/blob/master/src/modules/sensors/voted_sensors_update.cpp#L1253) in the Firmware.

matrix/AxisAngle.hpp
```diff
@@ -69,16 +69,11 @@ class AxisAngle : public Vector<Type, 3>
...
-            v(1) = ang*q(2)/mag;
-            v(2) = ang*q(3)/mag;
+        Type mag = Type(sqrt(q(1) * q(1) + q(2) * q(2) + q(3) * q(3)));
+        if (fabs(mag) >= Type(1e-10)) {
+            v = q.imag() * Type(Type(2) * atan2(mag, q(0)) / mag);
         } else {
-            v(0) = 0;
-            v(1) = 0;
-            v(2) = 0;
+            v = q.imag() * Type(Type(2) * float(sign(q(0))));
```
kamilritz:
```suggestion
            v = q.imag() * Type(Type(2) * Type(sign(q(0))));
```

matrix/AxisAngle.hpp
```diff
@@ -69,16 +69,11 @@ class AxisAngle : public Vector<Type, 3>
     AxisAngle(const Quaternion<Type> &q)
     {
         AxisAngle &v = *this;
-        Type ang = Type(2) * acos(q(0));
-        Type mag = sin(ang/2.0f);
-        if (fabs(mag) > 0) {
-            v(0) = ang*q(1)/mag;
-            v(1) = ang*q(2)/mag;
-            v(2) = ang*q(3)/mag;
+        Type mag = Type(sqrt(q(1) * q(1) + q(2) * q(2) + q(3) * q(3)));
```
dagar: Doesn't really matter, but might as well use `q.norm()`.

dagar: Sorry phone review, disregard.

dagar: Alternatively `q.imag().norm()`.

MaEtUgR: https://github.com/PX4/Matrix/pull/123

PR #122, opened 2020-03-03
Analytic inverse implementation

kamilritz:
This adds a template specification for the computation of the matrix inverse for the 2x2 and 3x3 case. The template specification is done in terms of the analytic expression. This could help to make the computation of the inverse in these case more efficient. In the worst case it for sure not less efficient and is also not adding to the size of an executable.


kamilritz:
@jkflying I added the error case. There is still the 9x9 matrix around line 80, that is a non-trivial case

PR #134, opened 2020-07-14
Switch operator() to return a const reference

MaEtUgR:
Proposed by @jkflying while he helped me debugging an independent issue.
I'm not entirely sure about the difference in practice. My code ran fine with and without this change.


MaEtUgR:
CI is just not reporting the status. Travis actually already succeeded if you open the link 😏

Issue #135, opened 2020-07-16
Dcm Type Error


gnos2010:
If the data type is not float, line 533 of the file Quaternion.hpp will appear a compile error.
If Change  ‘return Quaternion(Dcmf(dcm));’ to ‘return Quaternion(Dcm<Type>(dcm));’
It can adapt to more data types.

kamilritz:
@gnos2010 Thanks for raising this.

gnos2010:
I have been using this Matrix library to do some research work for some weeks. In order to improve the accuracy of numerical calculation, I use double precision floating-point data type. Thank you for your work.

gnos2010:
I sometimes use this Matrix library to do some research work, it is very convenient. TThank you for your work.

PR #137, opened 2020-07-17
Generalize float-only conversion

jkflying:
Fixes https://github.com/PX4/Matrix/issues/135


jkflying:
Fixed by the very speedy https://github.com/PX4/Matrix/pull/136

PR #140, opened 2020-08-03
SparseVector

kamilritz:
This PR adds a simple sparse vector class. This is useful for large vectors with only a few non zero entries. SparseVector is only storing the non-zero elements, but allows to access them with original index. Mapping the original index to the compressed index will happen at compile time.

In order for it to work properly c++14 is needed.

Additionally, this adds gtest. Current tests in matrix are getting quite confusing. 

@jkflying Thanks for your inputs.


matrix/SparseVector.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+public:
+    static constexpr int non_zeros() {
+        return N;
+    }
+    constexpr int index(int i) const {
+        return SparseVector::_indices[i];
+    }
+
+    SparseVector() {}
```
dagar, line 30:
```suggestion
    SparseVector() = default;
```

matrix/SparseVector.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+    SparseVector(const matrix::Vector<Type, N>& data) {
```
dagar, line 32:
```suggestion
    SparseVector(const matrix::Vector<Type, N>& data) : _data(data) {
```

matrix/SparseVector.hpp
kamilritz: Done

matrix/SparseVector.hpp
kamilritz: Done

matrix/SparseVector.hpp
```diff
@@ -0,0 +1,57 @@
...
+    static constexpr int non_zeros() {
+        return N;
+    }
+    constexpr int index(int i) const {
+        return SparseVector::_indices[i];
+    }
+
+    SparseVector() = default;
+
+    SparseVector(const matrix::Vector<Type, N>& data) : _data(data) {}
```
jkflying, line 32: I worry this could be confusing since it remaps the indices of the dense vector. But the other way would also be confusing. Perhaps it would be better to have named initializers to explain that it is doing this, instead of using an un-named constructor?

matrix/SparseVector.hpp
```diff
@@ -0,0 +1,57 @@
+#include "Vector.hpp"
+#include <cassert>
+
+template<int N> struct force_constexpr_eval {
+    static const int value = N;
+};
+
+// Vector that only store nonzero elements,
+// which indices are specified as parameter pack
+template<typename Type, int... Idxs>
```
jkflying, line 10: Right now there's actually nothing requiring that the indices are in ascending order. Maybe this is something that would be good to add? Maybe that is being overly restrictive?

matrix/SparseVector.hpp
kamilritz: Is it remapping indices? data is already of the size of the sparse vector.

matrix/SparseVector.hpp
kamilritz: At the moment I do not see any application where it would be needed to have it in an ascending order.

kamilritz:
> Unfortunately if it isn't in a template parameter then it won't guarantee to evaluate the recursive structure at compile time.

Ok, it should be possible to modify the auto code generation in ecl a bit, such that it uses at<i>().

matrix/SparseVector.hpp
jkflying: The existing behavior makes sense as long as this is just used for direct indexing, but looking forwards we will also want to do things like denseVector.dot(sparseVector), denseMatrix * sparseVector, denseVector += sparseVector, etc. These operations effectively treat the unpopulated indices as 0, but otherwise behave exactly like a dense vector with the same values in it.

Once we have these operators, I would expect `denseMatrix*denseVector` to give the same results as `sparseVector=denseVector; denseMatrix*sparseVector` assuming the denseVector only had non-zeros in the places the sparseVector could hold them.

But I also get that this might not be clear, and different people would have different opinions, so named initializers would make it obvious to anyone what is happening during an initialization. Something like `sparseVector.fromDense(denseVector)` and `sparseVector.initializeInternalRepresentation(denseVector)`

Some of this is coming from working extensively with Eigen's SparseMatrix / SparseVector structures. These are a bit different in that they detect and track sparsity dynamically, which won't work for us. But eventually we will want our SparseVector to be just as useful as Vector, and when that happens it should behave exactly like Vector except for keeping 0 implicitly for the unlisted indices.

kamilritz: Thanks for the explanation. This sounds great. Feel free to add it on top of this PR.

jkflying:
So the idea of the lookup-table based inverse seems to work, if maybe a big ugly:
https://godbolt.org/z/a9b6rh

matrix/SparseVector.hpp
jkflying: Will do

kamilritz:
@jkflying Thank you very much for your contribution and ideas. 
The new template parameter list  `SparseVector<Type, M, idxs...>`  is maybe a bit unclear to read in the client code, eg. `SparseVector<float, 7, 4, 6>`. When wrapping it with a type definition, it could get more clear: `SparseVector7f<4, 6>`. But I guess that is something the client should do.
I wonder if we want the client to specify the intended size M of the sparse vector at all. As you suggested, a specified M can be used to only allow operations with dense vectors of size M. But operations with any other dense vectors of size > M would also be possible. We could determine the max_index from idxs... and allow operations with any dense vector of size > max_index.

jkflying:
I had a long think about that, I 100% agree about making the interface worse. It is really that old argument of _explicit and less readable_, vs _implicit and more readable but needs more understanding of how it works_ . What finally convinced me to put the size in is to prevent bugs where we accidentally multiply `SparseVector<float, 1,2,3>` with a 6x8 matrix instead of the transposed 8x6 because we weren't sure if it was meant to be a 6-vector or 8-vector. If the signature was instead `SparseVector<float, 8, 1, 2, 3>` then it would give a compile time error doing the wrong thing. Maybe I'm overthinking this though :man_shrugging: 

Just to do a brain dump here, other functions which should be added are `norm()` and `longerThan()`,

kamilritz:
> I had a long think about that, I 100% agree about making the interface worse. It is really that old argument of _explicit and less readable_, vs _implicit and more readable but needs more understanding of how it works_ . What finally convinced me to put the size in is to prevent bugs where we accidentally multiply `SparseVector<float, 1,2,3>` with a 6x8 matrix instead of the transposed 8x6 because we weren't sure if it was meant to be a 6-vector or 8-vector. If the signature was instead `SparseVector<float, 8, 1, 2, 3>` then it would give a compile time error doing the wrong thing. Maybe I'm overthinking this though man_shrugging
> 
> Just to do a brain dump here, other functions which should be added are `norm()` and `longerThan()`,

I also thought a while about this, and I think I come to the same conclusion. It is probably rare that we want to change the underlying size of a sparse vector. But when this comes up, we still can have a separate constructor that only changes M, but keeps the rest the same.

Just wanted to be sure, that we move in the right direction.

kamilritz:
@jkflying found a nice solution to construct a sparse vector from either vector<M> or Type[N] array. I remove the fromDenseVector initialization as this can be achieved through slices and the new constructor.

jkflying:
Any idea why CI is failing?

kamilritz:
Probably the google test framework I added. I tried to fix it, but I am kind of stuck a bit.

kamilritz:
@jkflying CI is fixed now. I think this is ready to get merged. The google test framework will help to make other tests more readable as well.

PR #141, opened 2020-08-05
Construct Vector from Slice<1,M> alias row()

kamilritz:
Previously vectors could only be constructed from slice<M,1>.

### DIFF
--- a/matrix/Vector.hpp
+++ b/matrix/Vector.hpp
@@ -40,6 +40,15 @@ public:
     {
     }
 
+    template<size_t P, size_t Q, size_t DUMMY = 1>
+    Vector(const Slice<Type, 1, M, P, Q>& slice_in)
+    {
+        Vector &self(*this);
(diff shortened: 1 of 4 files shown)


matrix/Vector.hpp
```diff
@@ -40,6 +40,15 @@ class Vector : public Matrix<Type, M, 1>
     {
     }
 
+    template<size_t P, size_t Q, size_t DUMMY = 1>
```
kamilritz, line 43: Dummy parameter needed to avoid ambiguity for Vector<1,1>()

matrix/Vector.hpp
```diff
@@ -40,6 +40,15 @@ class Vector : public Matrix<Type, M, 1>
...
+    Vector(const Slice<Type, 1, M, P, Q>& slice_in)
+    {
```
kamilritz, line 45: Can not be constructed with any base class constructor.

PR #143, opened 2020-08-07
add possibility to iterate over SparseVector data at runtime

kamilritz:
We need a way to iterate over the data stored in the SparseVector. Since in a loop the index is not known at compile time, the current at<>() function can not be used. Add a runtime access version like at(i) that looks up the compressed index, is an expansive operation. Therefore I suggest we force the client to access the data directly via the compressed index. The actual index can be recovered with the index() function.

### DIFF
--- a/test/sparseVector.cpp
+++ b/test/sparseVector.cpp
@@ -35,9 +35,10 @@ TEST(sparseVectorTest, accessDataWithCompressedIndices) {
     const Vector3f vec(1.f, 2.f, 3.f);
     SparseVectorf<3, 0, 2> a(vec);
     for (size_t i = 0; i < a.non_zeros(); i++) {
-        a.atCompressedIndex(i) = 1.f;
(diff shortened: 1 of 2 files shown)


test/sparseVector.cpp
```diff
@@ -31,6 +31,15 @@ TEST(sparseVectorTest, initialisationFromVector) {
     EXPECT_FLOAT_EQ(a.at<2>(), vec(2));
 }
 
+TEST(sparseVectorTest, accessDataWithCompressedIndices) {
+    const Vector3f vec(1.f, 2.f, 3.f);
+    SparseVectorf<3, 0, 2> a(vec);
+    for (size_t i = 0; i < a.non_zeros(); i++) {
+        a.atCompressedIndex(i) = 1.f;
+        EXPECT_FLOAT_EQ(a.atCompressedIndex(i), 1.f);
```
jkflying, line 39: Better to do the checks using hardcoded at<>() after the loop, otherwise a misunderstanding in the indexing affects set and get the same.

PR #145, opened 2020-08-11
Make wrap() work with integer types

mortenfyhn:
I think wrap() in the helper functions fails in some corner cases when used with integer types.

Should it work with integer types? It's templated in such a way that you can certainly invoke it with ints, but it may not be intended for that. In any case, I have a fix, but I'm pushing some failing integer wrap tests first to see the CI fail.

If wrap() isn't meant to work with integer types, then should probably at least document it in the functions docstring,
or preferably use some clever template magic to make it not compile when Type is deduced to an integer type.

See the commit messages for more detail!

### DIFF
--- a/matrix/helper_functions.hpp
+++ b/matrix/helper_functions.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <type_traits>
 #include "math.hpp"
 
 #if defined (__PX4_NUTTX) || defined (__PX4_QURT)
@@ -40,14 +41,15 @@ bool isEqualF(const Type x, const Type y, const Type eps = 1e-4f)
 }
 
 /**
- * Wrap value to stay in range [low, high)
+ * Wrap floating point value to stay in range [low, high)
  *
  * @param x input possibly outside of the range
  * @param low lower limit of the allowed range
  * @param high upper limit of the allowed range
  * @return wrapped value inside the range
  */
-template<typename Type>
+template<typename Type,
+         std::enable_if_t<std::is_floating_point<Type>::value, int> = 0>
 Type wrap(Type x, Type low, Type high) {
     // already in range
     if (low <= x && x < high) {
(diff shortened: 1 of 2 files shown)


matrix/helper_functions.hpp
```diff
@@ -55,7 +55,7 @@ Type wrap(Type x, Type low, Type high) {
     }
 
     const Type range = high - low;
-    const Type inv_range = Type(1) / range; // should evaluate at compile time, multiplies below at runtime
+    const double inv_range = 1.0 / range; // should evaluate at compile time, multiplies below at runtime
```
jkflying, line 58: This could massively increase the cost of wrapping for `float`on Cortex-4 chips which don't support `double`. Could you check maybe with [godbolt](https://godbolt.org/) whether this still eliminates the temporaries for floats with ARM GCC 8? Otherwise we might need a template specialization for integer types.

mortenfyhn: Probably true, I'll have a look on godbolt tomorrow.

mortenfyhn:
Aight, I got this: https://godbolt.org/z/3YbzvE

I'm not super experienced with godbolt so I'm not sure I'm getting the comparison I need. Please let me know!

From what I can tell, my version invoked with floats runs 7 instructions more than the original, so it'll be a bit slower but not extremely much so. Do you agree? I still feel that maybe creating some specializations could be a good idea.

jkflying:
I made an example with wrap_pi: https://godbolt.org/z/svKsWf
The `_new` version has `dmul` instead of `fmul`, which will cause issues.

So this will need a specialization for integral types. Also, for integral types I don't think we'd want to do an 'invert and multiply' anyways, converting to floating point and back seems the wrong pattern because you will have imprecision at high values, which is fine for floats but for ints will be unexpected. Better to use the `%` operator for integers (and powers of 2 for the range when possible, for performance reasons).

mortenfyhn:
By the way: I just found out about the `-Wdouble-promotion` compiler warning, which seems to warn about exactly this issue. Maybe adding it to the build would be a good idea? If double promotion is unwanted for the entire repository, that is.

mortenfyhn:
I've split it into two functions now, with some funky business to make sure the right function is invoked for the right data types. I'm using `<type_traits>`, but that might not be acceptable in this repo?

jkflying:
Unfortunately PX4 doesn't have a lot of the more modern C++ stdlibs available on Nuttx, and <type_traits> is one of them (along with std::array, std::initializer_list etc, all of which should be fine in an embedded system). I think your best bet is to make the default call into an integer-optimized version, and have specializations for float and double, otherwise there are too many integer formats to specialize for.

mortenfyhn:
Got it, sorry for the delay here.

I've reworked it now. I put the shared floating point implementation into a sub-namespace "detail" to avoid duplicating it. Is that in line with the style you prefer here?

jkflying (approved):
Looks good, thanks for the contribution!

mortenfyhn:
Thanks for the merge!

PR #149, opened 2020-08-26
Sparse quadratic form

kamilritz:
Adds a method to compute x.T * A * x, where x is a SparseVector of size M and A is a square matrix of size MxM.


jkflying (approved):
Looks good, nice work!

PR #152, opened 2020-11-03
Precision of tol for float has been changed

Mitchell-Lee-93:
There were some cases that the results of pseudo inv becomes a null matrix when the input matrix is a floating-point number matrix such as follow: 
image
As a result, the actuators output unexpected values as follow:
image

I found out that the problem occurs as the precision of the tolerance for the float variables is too big, which is bigger than typical floating number precision(typically 7, reference: https://www.learncpp.com/cpp-tutorial/floating-point-numbers/).

So I simply decreased the precision for 2 digits and was able to obtain the following result:
image
and the typical actuator outputs log as follow:
image

### DIFF
--- a/matrix/PseudoInverse.hxx
+++ b/matrix/PseudoInverse.hxx
@@ -14,7 +14,7 @@
 template<typename Type>
 void fullRankCholeskyTolerance(Type &tol)
 {
-    tol /= 1000000000;
+    tol /= 10000000;
 }
 
 template<> inline


jkflying (approved):
From my side this looks good. Especially after some operations have been performed, 1e7 ratio between largest and smallest elements is very much at the limit with floats from my experience.

jlecoeur (approved):
Looks good to me

jlecoeur:
Should we add this to the tests?

jkflying:
I'll add to tests as a follow-up, this is exactly the kind of rank deficient matrix we need for them. It might also make sense to check what eg. Eigen, CHOLMOD etc are using for pivoting thresholds and rank determinance while we're at it. I remember Eigen using something like `max(abs(A)) * min(A.rows(), A.cols()) * ulp` which would be a little bit bigger still than what we have.

PR #153, opened 2020-11-04
Rework rank detection based on new test data

jkflying:
Inspired by the rank detection method in the Full Pivoting LU algorithm of [Eigen](https://eigen.tuxfamily.org/dox/FullPivLU_8h_source.html), I've reworked our rank detection for the pseudo-inverse. The new algorithm is:

```A.diag().max() * type_epsilon * N```
Which roughly translates as " Lose one ulp of precision per column in the matrix", which from an understanding of floating point numbers feels intuitively correct, as the error accumulates/propagates along the calculations with the 'bandwidth' relative to the largest magnitude element.

This makes it much more conservative than before, for example on the new test data the tolerance changed from 2.3e-8 to 2.1e-6, two orders of magnitude. Note also, that on the new test data the old rank detection was extremely marginal, as one of the pivots (which needed to be below the threshold for successful calculation) was 1.5e-8, well within numerical error relative to 2.3e-8.

### CODE REVIEW

File: matrix/PseudoInverse.hxx [hunk 1]
```diff
@@ -12,32 +12,20 @@
...
-    Vector<Type, N> d = A.diag();
-    Type tol = d.max();
-    for (size_t k = 0; k < N; k++) {
-        if ((d(k) > 0) && (d(k) < tol)) {
-            tol = d(k);
-        }
-    }
-    fullRankCholeskyTolerance<Type>(tol);
+    // Loses one ulp accuracy per row of diag, relative to largest magnitude
+    const Type tol = N * typeEpsilon<Type>() * A.diag().max();
```
jkflying, line 28: Matrix should be PSD so we don't need to worry about large negative numbers

PR #155, opened 2021-01-04
Use a single inverse implementation for a single matrix size

jkflying:
Previously, to handle all possible ranks of the matrix to be pseudo-inverted, a recursive template structure was used to have direct implementations of these. This has the negative side of significantly increasing flash size.

In order to prevent this, I've added the rank as an optional argument to the pure inverse function (which defaults to matrix size). This means that the detected rank from the cholesky decomposition can be passed directly to the inverse, which will then only try to invert that subset of the matrix, as if it was a smaller matrix. This removes the requirement to have implementations of other matrix sizes compiled in.

x86 unit test file sizes when compiled in MinSizeRel:
Before: pseudoInverse 116K, inverse 27K
After: pseudoInverse 36K, inverse 27K


LorenzMeier (approved):

jlecoeur (approved):
Great!

PR #161, opened 2021-04-13
Euler: simplify DCM constructor

dagar:
@MaEtUgR how about this instead?

### DIFF
--- a/matrix/Euler.hpp
+++ b/matrix/Euler.hpp
@@ -91,23 +91,20 @@ public:
     */
     Euler(const Dcm<Type> &dcm)
     {
-        Type phi_val = Type(atan2(dcm(2, 1), dcm(2, 2)));
-        Type theta_val = Type(asin(-dcm(2, 0)));
-        Type psi_val = Type(atan2(dcm(1, 0), dcm(0, 0)));
-        Type pi = Type(M_PI);
+        theta() = asin(-dcm(2, 0));
 
-        if (Type(fabs(theta_val - pi / Type(2))) < Type(1.0e-3)) {
-            phi_val = Type(0);
-            psi_val = Type(atan2(dcm(1, 2), dcm(0, 2)));
+        if ((fabs(theta() - Type(M_PI/2))) < Type(1.0e-3)) {
+            phi() = 0;
+            psi() = atan2(dcm(1, 2), dcm(0, 2));
 
-        } else if (Type(fabs(theta_val + pi / Type(2))) < Type(1.0e-3)) {
-            phi_val = Type(0);
-            psi_val = Type(atan2(-dcm(1, 2), -dcm(0, 2)));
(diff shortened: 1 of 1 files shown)


dagar:
Optimizes a bit better in PX4.

``` console
     VM SIZE    
 -------------- 
  -2.8%      -2    matrix::Dcm<float>::Dcm()
 -14.3%      -4    matrix::Vector<float, 2u>::norm() const
 -50.0%      -4    sensors::VehicleGPSPosition::PrintStatus()
  -3.1%      -6    matrix::Dcm<float>::Dcm(matrix::Quaternion<float> const&)
  -1.3%     -32    FixedwingAttitudeControl::Run()
 -17.0%     -32    matrix::Euler<float>::Euler(matrix::Dcm<float> const&)
 -26.1%     -72    matrix::Euler<float>::Euler(matrix::Quaternion<float> const&)
 -13.8%    -116    ControlMath::bodyzToAttitude(matrix::Vector3<float>, float, vehicle_attitude_setpoint_s&)
 -20.0%    -164    FixedwingPositionControl::vehicle_attitude_poll()
  -0.0%    -416    TOTAL
```

matrix/Euler.hpp
```diff
@@ -91,23 +91,20 @@ class Euler : public Vector<Type, 3>
...
-            phi_val = Type(0);
-            psi_val = Type(atan2(dcm(1, 2), dcm(0, 2)));
-
-        } else if (Type(fabs(theta_val + pi / Type(2))) < Type(1.0e-3)) {
-            phi_val = Type(0);
-            psi_val = Type(atan2(-dcm(1, 2), -dcm(0, 2)));
-        }
+        theta() = asin(-dcm(2, 0));
+
+        if ((fabs(theta() - Type(M_PI/2))) < Type(1.0e-3)) {
```
MaEtUgR, line 96:
```suggestion
        if ((fabs(theta() - Type(M_PI / 2))) < Type(1.0e-3)) {
```

matrix/Euler.hpp
```diff
@@ -91,23 +91,20 @@ class Euler : public Vector<Type, 3>
...
+            phi() = 0;
+            psi() = atan2(dcm(1, 2), dcm(0, 2));
 
-        phi() = phi_val;
-        theta() = theta_val;
-        psi() = psi_val;
+        } else if ((fabs(theta() + Type(M_PI/2))) < Type(1.0e-3)) {
```
MaEtUgR, line 100:
```suggestion
        } else if ((fabs(theta() + Type(M_PI / 2))) < Type(1.0e-3)) {
```

MaEtUgR (approved):
Yeah makes sense. I thought you might end up accidentally doing double calculations even in the float template if you don't explicitly cast. But otherwise it's a good simplification.

In terms of readability, it would make sense to have the else case first in some way because the corner cases are really rare corner cases that we just have to handle to not run into numerical issues in flight.

PR #162, opened 2021-04-13
Matrix: use naive per element copy instead of memcpy call

dagar:
This actually optimizes a bit better because the function call can be avoided entirely.

Saves about 7.3 kB of flash on px4_fmu-v5_default.

### DIFF
--- a/matrix/Matrix.hpp
+++ b/matrix/Matrix.hpp
@@ -103,7 +103,6 @@ public:
     {
         if (this != &other) {
             Matrix<Type, M, N> &self = *this;
-
             for (size_t i = 0; i < M; i++) {
                 for (size_t j = 0; j < N; j++) {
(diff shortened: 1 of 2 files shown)


matrix/Matrix.hpp
```diff
@@ -113,14 +102,25 @@ class Matrix
     Matrix<Type, M, N> & operator=(const Matrix<Type, M, N> &other)
     {
         if (this != &other) {
-            memcpy(_data, other._data, sizeof(_data));
+            Matrix<Type, M, N> &self = *this;
+
```
MaEtUgR, line 106:
```suggestion
```

MaEtUgR (approved):
Today I learned `memcpy` is not magically more efficient.
I looked through the changes and they all make sense after you explained how it's useful for optimization to explicitly do the assignments.

PR #164, opened 2021-07-09
quaternion exponential

romain-chiap:
I added a static method to compute the quaternion exponential for a 3d vector u.

This can be used to update a quaternion kinematics from the body rates
rather than using
`qk+1=qk+qk.derivative1(wb)*dt`
we can use
`qk+1=qk*expq(dt*wb/2)`
which is a more robust update.
A re-normalization step is necessary with the first method, but not the second one.

I also created a check in test/attitude.cpp, which passed successfully.
I did not format with Astyle, as this is not installed on my machine.


bresch:
Great, thanks, we should try it in EKF2 now.

matrix/Quaternion.hpp
```diff
@@ -325,6 +325,36 @@ class Quaternion : public Vector<Type, 4>
         return v * q  * Type(0.5);
     }
 
+    /**
+     * Computes the quaternion exponential of the 3D vector u
+     * return a quaternion computed as
+     * expq(u)=[cos||u||, sinc||u||*u
```
bresch, line 331:
```suggestion
     * expq(u)=[cos||u||, sinc||u||*u]
```

bresch (changes requested):
Could you please add the reference to the paper you mentioned in Slack? (https://arc.aiaa.org/doi/pdf/10.2514/1.G004578)

romain-chiap:
> Great, thanks, we should try it in EKF2 now.

I first wanted to use it in sih, because the quaternion update with addition gave me problems when I implemented a fixed wing spiralling down.
Equations 24 and 25 in the paper could also be used for a higher order integration method, like RK4. Eventually I'll implement that.

romain-chiap:
let me create a test for the inverse right Jacobean before merging.

matrix/Quaternion.hpp
```diff
@@ -325,6 +325,64 @@ class Quaternion : public Vector<Type, 4>
...
+      */
+	static Quaternion expq(const Vector3<Type> u)
+	{
+		const Type tol = Type(1.0e-6);
+		Type u_norm = u.norm();
+		Type sinc;
+		if (u_norm < tol) {
+			sinc = Type(1.0) - u_norm * u_norm / Type(6.0); // this ensure an error smaller than tol^3
+		} else {
+			sinc = Type(sin(u_norm) / u_norm);
```
bresch, line 356: If this is supposed to run at high speed, should we maybe consider its Taylor series? 
DeepinScreenshot_select-area_20210726173609

romain-chiap: Isn't it what's implemented already?
in PX4-Autopilot/platforms/nuttx/NuttX/nuttx/libs/libc/math/[lib_sin.c](https://github.com/amlweems/wiced-history/blob/master/WICED/RTOS/NuttX/ver7.8/libc/math/lib_sin.c)

dagar: FYI not that the distinction really matters here, but we use libm from the toolchain (newlib) and not what's directly in NuttX.

https://github.com/eblot/newlib/blob/master/newlib/libm/math/kf_sin.c

bresch: > Isn't it what's implemented already?
in PX4-Autopilot/platforms/nuttx/NuttX/nuttx/libs/libc/math/lib_sin.c

Yes, but that's just for `sin(x)`; you then need to divide by `x` again. Using the Taylor series for `sin(x)/x` directly removes the need for any division (and you maybe don't even need to compute the 10 first elements?).

romain-chiap: ok, here is the idea:

if norm(u) is smaller than 0.2, we compute the 4 first terms of the Taylor series, which ensures an error smaller than 10^-10 for double (for float, the machine error is 10^-8 so the error is smaller than the machine error)
if norm(u) is bigger than 0.2, we call sin(u)/u and cos(u)

since the update is 
`qk+1=qk*expq(dt*wb/2)`
with dt about 4*10-3, wb can be as big as 100 rad/s in norm!

In practice such high body rates never happens, however I want to keep the correct formulation for higher norms just in case.

romain-chiap:
I want to use expq() in [PR-17991](https://github.com/PX4/PX4-Autopilot/pull/17991)

matrix/Quaternion.hpp
bresch: Cool, I like it!

bresch (changes requested):
Looks good to me now, just apply this formatting and I think it's ready to go: https://github.com/romain-chiap/PX4-Matrix/pull/1

bresch (approved):

dagar (approved):

matrix/Quaternion.hpp
```diff
@@ -326,34 +326,60 @@ class Quaternion : public Vector<Type, 4>
...
+      * rather than using
+      * qk+1=qk+qk.derivative1(wb)*dt
+      * we can use
+      * qk+1=qk*expq(dt*wb/2)
+      * which is a more robust update.
+      * A re-normalization step might necessary with both methods.
+      *
+      * @param u 3D vector u
+      */
+	static Quaternion expq(const Vector3<Type> u)
```
jgoppert: For integration of rigid body dynamics, we may want to also add the SE3 rigid body translation/rotation matrix exponential and logarithm. You can find a good summary here. https://ethaneade.com/lie.pdf This would improve the integration accuracy for the x, y, z positions as well as the attitude. I would recommend considering this for the propagation step for EKF2.

PR #165, opened 2021-08-31
matrix inversion: improve memory usage & runtime

bkueng:
- use associativity of matrix operations to reduce size of temporary matrices and number of multiplications in the M <= N case
- minimize the number of temporary matrices required

In particular this reduces stack usage for the PX4 control allocator module by 1.3KB (1.2KB for 1. commit, ~150B for 2. commit).

Runtime:
||master (Os) |this PR (Os)  | this PR (O2)| this PR (O3)|
--- | --- | ---|---|---
|F4|0.45ms|0.3ms|0.3ms|0.12ms (+7KB flash size)|
|F7||0.12ms||0.07ms|


dagar:
CI is broken (travis-ci free shutdown).

dagar (approved):

matrix/PseudoInverse.hpp
```diff
@@ -21,31 +21,38 @@ namespace matrix
...
-Matrix<Type, N, M> geninv(const Matrix<Type, M, N> & G)
+bool geninv(const Matrix<Type, M, N> & G, Matrix<Type, N, M>& res)
 {
     size_t rank;
     if (M <= N) {
         SquareMatrix<Type, M> A = G * G.transpose();
         SquareMatrix<Type, M> L = fullRankCholesky(A, rank);
 
-        SquareMatrix<Type, M> LtL = L.transpose() * L;
+        A = L.transpose() * L;
```
jkflying, line 31: So, this renaming shouldn't actually affect anything performance wise (and IMO makea the algorithm harder to understand), since the first thing the compiler does anyway is transform the code back into a SSA representation anyways.

jkflying (commented):
Good catch!

matrix/PseudoInverse.hpp
jkflying: Interesting, maybe there were some side-effects in the assignment constructor or the stack address offsets from the extra structure are somehow important.

PR #167, opened 2021-10-06
slice: add min and max functions

bresch:
Needed for the PX4 dynamic control allocation project.


dagar (approved):

bresch:
> Add a quick test?

Done

